PR #159, opened 2024-07-22
feat: temporary localnet config

neilcampbell:
**Features**
- Allow temporary customisation of the localnet config using url search params. This is used when launching lora via `algokit explore localnet` inside a GitHub codespace.

**Fixes**
- Set selected network to localnet when deleting a custom network that is currently selected.
- Reset data provider when resetting the settings of a default network that has been modified and is currently selected.
- Disallow creating custom networks with the name `settings` or `app-studio`, as they collide with app routes and will cause weird behaviours.
- Add id to multiselect so the label is valid.

### DIFF
--- a/src/features/network/pages/network-page.tsx
+++ b/src/features/network/pages/network-page.tsx
@@ -1,7 +1,7 @@
 import { useRequiredParam } from '@/features/common/hooks/use-required-param'
 import { UrlParams } from '@/routes/urls'
-import { temporaryLocalNetSearchParams, useNetworkConfigs, useSelectedNetwork } from '@/features/network/data'
-import { useLocation, useNavigate, useSearchParams } from 'react-router-dom'
+import { useNetworkConfigs, useSelectedNetwork } from '@/features/network/data'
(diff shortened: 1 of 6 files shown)


src/features/network/pages/network-page.tsx
```diff
@@ -26,5 +28,21 @@ export function NetworkPage({ children }: Props) {
     }
   }, [hash, pathname, search, navigate, selectedNetwork, isWildcardNetworkRoute])
 
+  useEffect(() => {
```
PatrickDinh, line 31: I'm wondering if we should move this outside of a react component. My reasoning is that the logic to read the params is outside of a React component => the logic to delete them should be outside too.

PatrickDinh (approved):

src/features/network/pages/network-page.tsx
neilcampbell: We could extract into a hook, but it needs to needs to still run as an effect, as it uses react router to do the delete.

Issue #158, opened 2024-07-22
ARC69 + ARC3 Asset isn't correctly displaying the image


neilcampbell:
https://lora.algokit.io/mainnet/asset/909935715

Also add as a test case.

PR #161, opened 2024-07-23
fix: handle an asset which uses both arc-3 and arc-69

neilcampbell:
Fixes https://github.com/algorandfoundation/algokit-lora/issues/158

### DIFF
--- a/src/features/assets/data/asset-metadata.ts
+++ b/src/features/assets/data/asset-metadata.ts
@@ -1,7 +1,7 @@
 import { TransactionResult, TransactionSearchResults } from '@algorandfoundation/algokit-utils/types/indexer'
 import { flattenTransactionResult } from '@/features/transactions/utils/flatten-transaction-result'
 import { TransactionType } from 'algosdk'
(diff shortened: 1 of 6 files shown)


src/features/assets/data/asset-metadata.ts
```diff
@@ -15,20 +15,24 @@ import { replaceIpfsWithGatewayIfNeeded } from '../utils/replace-ipfs-with-gatew
 // ARCs are community standard, therefore, there are edge cases
 // For example:
 // - An asset can follow ARC-69 and ARC-19 at the same time: https://allo.info/asset/1559471783/nft
+// - An asset can follow ARC-69 and ARC-3 at the same time: https://allo.info/asset/909935715/nft
 // - An asset can follow ARC-3 and ARC-19 at the same time: https://allo.info/asset/1494117806/nft
+
```
PatrickDinh, line 20: We don't need the extra line here, I think

PatrickDinh (approved):

PR #174, opened 2024-08-01
Decode ARC-32 app call transactions

PatrickDinh:
Feat: decode ARC-32 app call transactions
- A button to allow users upload ARC-32 app spec to any on-chain application. This button is hidden until the entire app studio feature is ready. The users are always able to remove the hidden CSS class to view the button.
- When an app has ARC-32 app spec attached. App call transactions are decoded to show:
  - The method name. 
  - The method arguments.
  - The return value.


src/tests/object-mother/transaction-result.ts
```diff
@@ -3047,4 +3047,298 @@ export const transactionResultMother = {
       'tx-type': TransactionType.axfer,
     })
   },
+  ['testnet-QY4K4IC2Z5RQ5OM2LHZH7UAFJJ44VUDSVOIAI67LMVTU4BHODP5A']: () => {
```
PatrickDinh, line 3050: Should we break this file up?

neilcampbell: We could. Potentially we could store each transaction as a ts file. This would then simply expose those files to the tests.

src/features/abi-methods/data/types/index.ts
```diff
@@ -0,0 +1,11 @@
+import { AlgoAppSpec as Arc32AppSpec } from '@/features/abi-methods/data/types/arc-32/application'
+
+// Only ARC-32 is supported for now
```
neilcampbell, line 3: FYI we also need to support ARC-4. There is a chat in slack about it.

src/features/abi-methods/models/index.ts
```diff
@@ -0,0 +1,54 @@
+export enum AbiType {
+  String = 'string',
```
neilcampbell, line 2: We mostly use PascalCase for the enum values elsewhere. Do the values need to be camel/lowercase?

src/features/accounts/data/account-result.ts
```diff
@@ -31,36 +30,38 @@ const getAccountResult = async (address: Address) => {
   }
 }
 
-const syncAssociatedDataAndReturnAccountResultAtom = atom(null, async (get, set, address: Address) => {
-  const accountResult = await getAccountResult(address)
-  const assetResults = get(assetResultsAtom)
-  const applicationResults = get(applicationResultsAtom)
+const syncAssociatedDataAndReturnAccountResultAtom = atom(null, (get, set, address: Address) => {
+  return atom(async () => {
```
neilcampbell, line 34: To keep this close to the previous behaviour, can we keep the write only atom as async and perform all evaluations inside that. We can then make the returned atom a simple `atom(() => accountResult}`

src/features/assets/data/asset-metadata.ts
```diff
@@ -150,7 +151,13 @@ const getAssetMetadataResult = async (assetResult: AssetResult) => {
   return await createAssetMetadataResult(assetResult, assetConfigTransactionResults[0])
 }
 
+const assetMetadataResultAtomBuilder = (assetResult: AssetResult) => {
```
neilcampbell, line 154: We could write a generic function for this an export from atoms-in-atom.

src/features/blocks/data/block-result.ts
```diff
@@ -111,12 +110,14 @@ export const addStateExtractedFromBlocksAtom = atom(
   }
 )
 
-const syncAssociatedDataAndReturnBlockResultAtom = atom(null, async (_get, set, round: Round) => {
-  const [blockResult, transactionResults, groupResults] = await getBlockAndExtractData(round)
+const syncAssociatedDataAndReturnBlockResultAtom = atom(null, (_get, set, round: Round) => {
+  return atom(async () => {
```
neilcampbell, line 114: Same as previous comment relating to the same thing.

src/features/common/data/atoms-in-atom.ts
```diff
@@ -1,12 +1,11 @@
 import { WritableAtom, atom, type Atom } from 'jotai'
 import { dataStore } from './data-store'
 
-const defaultCreateInitialValueAtom = <T>(value: T) => atom(() => value)
-
 export const createTimestamp = (): number => Date.now()
 
 export const createAtomAndTimestamp = <T>(value: T) => {
-  return [defaultCreateInitialValueAtom(value), createTimestamp()] as const
+  const createAtomFunc = (value: T) => atom(() => Promise.resolve(value))
```
neilcampbell, line 7: Probably no need to have this defined as a function anymore. It can be inlined. It's also interesting that we need to do a Promise.resolve now. Is that needed?

src/features/common/data/state-cleanup.ts
```diff
@@ -32,13 +33,14 @@ const stateCleanupEffect = atomEffect((get, set) => {
     removeExpired(applicationResultsAtom)
     removeExpired(assetMetadataResultsAtom)
     removeExpired(assetResultsAtom)
+    removeExpired(applicationsAppSpecsAtom)
```
neilcampbell, line 36: Assume you've tested that once an app spec has been evicted, a fetch to the DB occurs again?

src/features/explore/pages/explore-page.test.tsx
```diff
@@ -120,8 +120,8 @@ describe('explore-page', () => {
       },
       {
         syncedRound: 0,
-        blocks: new Map<Round, readonly [Atom<BlockResult>, number]>(),
-        transactions: new Map<TransactionId, readonly [Atom<TransactionResult>, number]>(),
+        blocks: new Map<Round, readonly [Atom<Promise<BlockResult>>, number]>(),
```
neilcampbell, line 123: Seems like this PR has remove the concept of AsyncMaybe?

src/features/groups/data/group-result.ts
```diff
@@ -5,17 +5,19 @@ import { addStateExtractedFromBlocksAtom, getBlockAndExtractData } from '@/featu
 import { invariant } from '@/utils/invariant'
 import { atomsInAtom } from '@/features/common/data'
 
-const syncAssociatedDataAndReturnGroupResultAtom = atom(null, async (_get, set, groupId: GroupId, round: Round) => {
-  const [blockResult, transactionResults, groupResults] = await getBlockAndExtractData(round)
+const syncAssociatedDataAndReturnGroupResultAtom = atom(null, (_get, set, groupId: GroupId, round: Round) => {
```
neilcampbell, line 8: Same as other related comment

src/features/transactions-graph/components/app-call-transaction-tooltip-content.tsx
```diff
@@ -2,40 +2,63 @@ import { AppCallTransaction, InnerAppCallTransaction } from '@/features/transact
...
+        },
+        {
+          dt: applicationIdLabel,
+          dd: <ApplicationLink applicationId={transaction.applicationId} />,
+        },
+        {
+          dt: transactionFeeLabel,
+          dd: <DisplayAlgo amount={transaction.fee} />,
+        },
+      ].filter(isDefined),
```
neilcampbell, line 60: Don't think this is needed.

src/features/transactions-graph/components/app-call-transaction-tooltip-content.tsx
```diff
@@ -2,40 +2,63 @@ import { AppCallTransaction, InnerAppCallTransaction } from '@/features/transact
...
+import { useAtomValue } from 'jotai/index'
+import { loadable } from 'jotai/utils'
+import { RenderLoadable } from '@/features/common/components/render-loadable'
+import { AbiMethod } from '@/features/abi-methods/models'
 
 export function AppCallTransactionTooltipContent({ transaction }: { transaction: AppCallTransaction | InnerAppCallTransaction }) {
+  const loadableAbiMethod = useAtomValue(loadable(transaction.abiMethod))
+
+  return (
+    <RenderLoadable loadable={loadableAbiMethod}>
```
neilcampbell, line 23: Same comment about the application call info component loader granularity.

src/features/transactions-graph/components/horizontal.tsx
```diff
@@ -83,11 +84,25 @@ function VectorLabelText({ type }: { type: LabelType }) {
   return undefined
 }
 
+function AppCallAbiMethodName({ transaction }: { transaction: AppCallTransaction | InnerAppCallTransaction }) {
+  return (
+    <RenderInlineAsyncAtom atom={transaction.abiMethod}>
```
neilcampbell, line 89: For this one, we could potentially render the loading state and nothing, as the majority of circumstances we won't have app specs. Thoughts?

src/features/transactions/data/transaction.ts
```diff
@@ -6,19 +6,20 @@ import { TransactionId } from './types'
 import { asTransaction } from '../mappers/transaction-mappers'
 import { getTransactionResultAtom } from './transaction-result'
 import { assetSummaryResolver } from '@/features/assets/data/asset-summary'
+import { abiMethodResolver } from '@/features/abi-methods/data/abi-method'
 
 export const createTransactionsAtom = (transactionResults: TransactionResult[]) => {
-  return atom((_get) => {
+  return atom((_) => {
```
neilcampbell, line 12:
```suggestion
  return atom(() => {
```

src/utils/sleep.ts
```diff
@@ -0,0 +1 @@
+export const sleep = (ms: number) => new Promise((resolve) => setTimeout(resolve, ms))
```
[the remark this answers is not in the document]
PatrickDinh: TIL about the setTimeOut in node

src/features/transactions/components/app-call-transaction-info.tsx
```diff
@@ -21,92 +27,133 @@ const foreignApplicationsTabId = 'foreign-applications'
...
+  transaction,
+  abiMethod,
+}: {
+  transaction: AppCallTransaction | InnerAppCallTransaction
+  abiMethod: AbiMethod | undefined
+}) {
+  const tabs = useMemo(
+    () =>
+      [
+        abiMethod
```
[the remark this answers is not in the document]
PatrickDinh: I changed it to spread operator but to be honest I keep switching because it and return/filter undefined. Do you have any reasons why we should use one over another?

src/features/abi-methods/components/decoded-abi-method-arguments.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+export function DecodedAbiMethodArguments({ method }: { method: AbiMethod }) {
+  const components = useMemo(
+    () =>
+      method.arguments.map((argument) => {
+        if (argument.type === AbiType.Transaction) {
+          return (
+            <>
+              <span>{argument.name}: </span>
+              <TransactionLink className={'text-primary underline'} transactionId={argument.value}>
```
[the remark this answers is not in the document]
PatrickDinh: >Also className={'text-primary underline'} can be className="text-primary underline". Do you prefer one over the other? I've notice it's pretty mixed in the codebase.
I don't have a preference, I tried to className=" but sometimes AI generated code keep inserting className={ and I forgot to switch it back

src/features/abi-methods/components/decoded-abi-method-arguments.tsx
```diff
@@ -0,0 +1,74 @@
...
+        )
+      }),
+    [method.arguments]
+  )
+
+  return (
+    <ul className={'pl-4'}>
+      {components.map((component, index, arr) => (
+        <li key={index}>
+          <>
```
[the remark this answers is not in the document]
PatrickDinh: yes we don't

src/features/transactions-graph/components/horizontal.tsx
PatrickDinh: As discussed, I leave this unchanged.

src/features/abi-methods/components/decoded-abi-method.tsx
```diff
@@ -0,0 +1,25 @@
...
+  return (
+    <div className="max-h-[450px] overflow-x-scroll">
+      <div>
+        <span>{abiMethod.name}(</span>
+        <DecodedAbiMethodArguments method={abiMethod} />)
+      </div>
+      <div className={'my-4'} />
+      <div>
+        <span>Returns: </span>
+        <div className="inline">
```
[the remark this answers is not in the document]
PatrickDinh: I removed the inline div because it wasn't needed. I didn't move `DecodedAbiMethodReturnValue` inside the `span` because `DecodedAbiMethodReturnValue` may contain div and ul

src/features/abi-methods/data/abi-method.ts
```diff
@@ -0,0 +1,255 @@
...
+      name: abiMethod.name,
+      arguments: methodArguments,
+      return: methodReturn,
+    } satisfies AbiMethod
+  })
+}
+
+const getAbiMethodAtom = (transaction: TransactionResult): Atom<Promise<algosdk.ABIMethod | undefined>> => {
+  return atom(async (get) => {
+    if (!isValidAppCallTransaction(transaction)) {
```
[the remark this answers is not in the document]
PatrickDinh: I tend to keep the guard on top the function regardless if the caller already check it so that if this function (`getAbiMethodAtom`) is called somewhere, the guard makes sure that it behaves as expected.

src/features/abi-methods/data/index.ts
```diff
@@ -0,0 +1,93 @@
...
+  (applicationId) => applicationId
+)
+
+export const useSetAppSpec = (applicationId: ApplicationId) => {
+  const setAppSpec = useSetAtom(getApplicationsAppSpecsAtom(applicationId))
+
+  return useCallback(
+    async ({
+      standard,
+      validFromRound,
```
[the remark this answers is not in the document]
PatrickDinh: From the Algorand REST API docs:
> first-valid required	[fv] First valid round for this transaction.	integer
I think this is inclusive

src/features/abi-methods/data/test-app-specs/auction.arc32.json
```diff
@@ -0,0 +1,164 @@
+{
```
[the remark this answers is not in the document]
PatrickDinh: I moved them to `src/tests/test-app-specs`. I confirmed that they aren't included in the build output.

src/features/abi-methods/data/index.ts
```diff
@@ -0,0 +1,104 @@
...
+      if (!value.length) {
+        return getAppSpecs(applicationId)
+      }
+      return value
+    },
+    async (_, set, appSpecs: AppSpecVersion[]) => {
+      await writeAppSpecs(applicationId, appSpecs)
+      set(appSpecVersions, appSpecs)
+    }
+  )
```
neilcampbell, line 37: This will result in use going to indexed db each time when we don't have any app specs.
We only need to do that on the first call.

To solve this we could have an `undefined` state for `appSpecVersions` and use an effect to set that atom to empty when we read and no results come back from the db and the previous value is undefined.

src/features/abi-methods/data/abi-method.ts
```diff
@@ -0,0 +1,255 @@
...
+    value: Number(abiValue),
+  }
+}
+
+const isValidAppCallTransaction = (transaction: TransactionResult): boolean => {
+  return (
+    transaction['tx-type'] === TransactionType.appl &&
+    Boolean(transaction['application-transaction']) &&
+    Boolean(transaction['confirmed-round']) &&
+    Boolean(transaction['application-transaction']?.['application-args']) &&
```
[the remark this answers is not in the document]
neilcampbell: I was thinking we could decode to bytes and see if it's the correct number of bytes to potentially be a method selector. If it's not then we know for sure it's not an ABI call.

src/features/abi-methods/data/index.ts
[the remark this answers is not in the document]
neilcampbell: Potentially something that may not have come across in that comment is that I think we should match the naming too (and semantics). Thoughts?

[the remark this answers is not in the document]
PatrickDinh: yes, I have updated them to `roundFirstValid` and `roundLastValid`

src/features/abi-methods/data/abi-method.ts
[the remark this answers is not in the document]
PatrickDinh: As discussed, I removed the guards and use invariants

neilcampbell (approved):

Issue #177, opened 2024-08-07
[local-dev-experience-ui] Users can fund an existing LocalNet account in lora


neilcampbell:
## Description

As an AlgoKit user
I want to easily fund an existing LocalNet account
So that I have enough ALGO to use my account on the network

## Acceptance Criteria
- Upon entering a valid address and amount, the account is funded with the amount specified using the KMD dispenser account

Issue #176, opened 2024-08-07
[local-dev-experience-ui] Users can create a funded LocalNet account for use in lora


neilcampbell:
## Description

As an AlgoKit user
I want to easily create a funded LocalNet account
So that I can connect that account and interact with LocalNet from within lora

## Acceptance Criteria
- A lora dev wallet is create inside KMD
- A new account is created inside the lora dev wallet in KMD
- The new account if funded ALGO using the KMD dispenser account
- The lora dev wallet in KMD can be chosen when connecting a wallet inside lora
- The new account can be selected once the KMD lora dev wallet has been connected

PR #184, opened 2024-08-07
feat: LocalNet funding

neilcampbell:
Completes #176 and #177

- Add support for creating and then funding a new LocalNet account via KMD
- Add support for funding an existing LocalNet account
- A few small cleanup and refactors I noticed along the way

Note: I'll add some tests once we have TestNet funding support, so I can think more holistically about it.

fund-localnet


src/features/fund/components/fund-account-form.tsx
```diff
@@ -0,0 +1,74 @@
...
+const amountLabel = 'Amount'
+
+const fundFormSchema = zfd.formData({
+  receiver: zfd.text().optional(),
+  amount: zfd.numeric(z.number({ required_error: 'Required', invalid_type_error: 'Required' }).min(0)),
+})
+
+type Props = {
+  onCreateReceiver?: () => Promise<Address>
+  onFund: (receiver: Address, amount: AlgoAmount) => Promise<void>
```
PatrickDinh, line 25: maybe we should rename `onFund` to `onFunded`

PatrickDinh (approved):

src/features/fund/components/fund-account-form.tsx
neilcampbell: I've changed to onSubmit, which I think is clearer.

Issue #180, opened 2024-08-07
[local-dev-experience-ui] Users can fund an existing TestNet account in lora


neilcampbell:
## Description

As an AlgoKit user
I want to easily fund an existing TestNet account
So that I have enough ALGO to use my account on the network

## Acceptance Criteria
- Users can login to access the TestNet dispenser API
- Upon entering a valid address and amount, the account is funded with the amount specified using the TestNet dispenser API

## Tasks
- [x] Update the Auth0 email process to use Amazon SES @neilcampbell
- [x] Update any relevant packages in the dispenser API and redeploy the stack @neilcampbell 
- [x] Build the feature @neilcampbell

PR #191, opened 2024-08-13
feat: fund on testnet

neilcampbell:
- Completes #180

testnet_funding


src/features/forms/components/number-form-item.tsx
```diff
@@ -42,15 +42,17 @@ export interface NumberFormItemProps<TSchema extends Record<string, unknown> = R
   decimalScale?: number
   thousandSeparator?: boolean
   placeholder?: string
+  helpText?: string | ReactElement
```
neilcampbell, line 45: FYI. I haven't propagated helpText to all the form items, as I was trying to constrain the number of components I touched. I can do this or we can do it as we use it?

src/features/forms/components/validation-error-or-help-message.tsx
PatrickDinh, line 1: I think we should split this into 2 components, there will be scenarios when we want to display error message and help text.

src/features/fund/fund-page.test.tsx
```diff
@@ -0,0 +1,131 @@
...
+
+      return executeComponentTest(
+        () =>
+          render(
+            <Auth0Provider domain="test" clientId="test">
+              <FundPage />
+            </Auth0Provider>
+          ),
+        async (component) => {
+          expect(1).toBe(1)
```
PatrickDinh, line 106: do we need this?

.github/actions/static-site-transforms/action.yaml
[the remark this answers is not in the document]
neilcampbell: 100%, it was lifted from there. For some reason I was thinking that only workflows are reuse-able. Will update.

src/features/common/components/render-inline-async-atom.tsx
```diff
@@ -8,5 +8,12 @@ const transformError = (error: Error) => {
 }
 
 export function RenderInlineAsyncAtom<T>(props: RenderAsyncAtomProps<T>) {
-  return <RenderAsyncAtom transformError={transformError} fallback={<Ellipsis className="size-4 h-full animate-pulse" />} {...props} />
+  const { fallback, ...rest } = props
```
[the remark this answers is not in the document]
neilcampbell: That's fair, I'll change to now allow fallback to be passed at all then.

src/features/forms/components/validation-error-or-help-message.tsx
neilcampbell: We can. I went back and forth, but I think we actually want to display help text or error message, which is why I set up like this. If we change to a different behaviour we can split.

PatrickDinh (approved):

PR #228, opened 2024-08-19
feat: create app interface from app lab page

PatrickDinh:
https://github.com/user-attachments/assets/8ab2cd39-49bc-47a4-8eea-b78e84efea60


src/features/common/data/indexed-db.ts
```diff
@@ -1,32 +1,89 @@
...
 }
 
 const dbMigrations = [
-  (db: IDBPDatabase<LoraDBSchema>) => {
-    db.createObjectStore('applications-app-specs')
+  async (db: IDBPDatabase<DBSchema>) => {
+    const v1Db = db as unknown as IDBPDatabase<LoraDBSchemaV1>
+    v1Db.createObjectStore('applications-app-specs')
+  },
+  async (db: IDBPDatabase<DBSchema>, transaction: IDBPTransaction<DBSchema, StoreNames<DBSchema>[], 'versionchange'>) => {
```
neilcampbell, line 56: It's good to have run a migration, however I don't think we need to do do it at this stage. Feels like it'll keep the code simpler to just make the breaking change?

We can probably apply a run that until we release this publicly, we are ok to simply make breaking schema changes. Thoughts?

Alternatively we can apply schema migrations and just clean them up prior to releasing?

src/features/app-studio/pages/app-studio-wip-page.tsx
```diff
@@ -0,0 +1,23 @@
...
+
+export function AppStudioWipPage() {
+  return (
+    <>
+      <PageTitle title={appStudioPageTitle} />
+      <div className="grid grid-cols-3 gap-4">
+        <Card>
+          <CardContent className="relative min-h-44 space-y-1">
+            <h2>{newContractLabel}</h2>
+            <p>Upload an ARC-32 JSON file</p>
```
neilcampbell, line 16: I think there is a bit of terminology we want to refine. Let's do this as part of naming.

neilcampbell (approved):

src/features/common/data/indexed-db.ts
PatrickDinh: I don't mind applying breaking changes 😄 . I think the migration here is good for establishing a pattern.

neilcampbell: Yeah agreed. Maybe let's run the migrations as a way of formalising the process and then cleanup any migration scripts prior to the public release, so we don't have that additional code.

neilcampbell: Also just realised the db and table is created by default, so we'll need to make sure we account for that.

src/features/app-interfaces/data/index.ts
```diff
@@ -0,0 +1,95 @@
...
+    }
+  )
+}
+
+export const [appInterfacesAtom, getAppInterfaceAtom] = writableAtomCache(
+  createWritableAppInterfaceEntityAtom,
+  (applicationId: ApplicationId) => applicationId
+)
+
+export const useCreateAppInterface = () => {
```
PatrickDinh, line 40: @neilcampbell I refactored this method to only for creating an app interface because I think that we won't need the logic for updating for a while. This will keep the code a bit simpler. I'll add the update logic as a separated method later on.

neilcampbell (approved):

PR #232, opened 2024-08-21
feat: use network genesis hash in the db name

PatrickDinh:
https://github.com/user-attachments/assets/8d905567-12ba-4bc4-8ced-5c8be456ea15

### DIFF
--- a/src/features/common/data/indexed-db.ts
+++ b/src/features/common/data/indexed-db.ts
@@ -4,6 +4,7 @@ import { ApplicationId } from '@/features/applications/data/types'
 import { genesisHashAtom } from '@/features/blocks/data'
 import { atom } from 'jotai/index'
 import { selectedNetworkAtomId } from '@/features/network/data'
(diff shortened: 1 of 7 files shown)


src/features/common/data/indexed-db.ts
```diff
@@ -24,7 +26,8 @@ interface LoraDBSchemaV2 extends DBSchema {
 }
 
 export async function getDbConnection(networkId: string) {
-  return await openDB<LoraDBSchemaV2>(networkId, 2, {
+  const genesisHash = await dataStore.get(genesisHashAtom)
```
neilcampbell, line 29: Accessing like this means that this function will never be updated if the atom is updated. It means that this function relies on the behaviour of the provider to get the correct value.

I'll push a change which matches what we do for networkId.

neilcampbell: I can't think of any other options, so I think this is fine.

PR #246, opened 2024-08-30
feat: build pay transactions using transaction wizard

neilcampbell:
- Added support for building pay transaction using the transaction wizard
- Made a few small fixes I noticed along the way

transaction_wizard


src/features/transaction-wizard/data/payment-transactions.ts
```diff
@@ -0,0 +1,143 @@
...
+      ? {
+          firstValidRound: data.validRounds.firstValid,
+          lastValidRound: data.validRounds.lastValid,
+        }
+      : undefined),
+  })
+  return transaction
+}
+
+const createAccountClose = async (data: z.infer<typeof accountCloseSchema>) => {
```
neilcampbell, line 49: There is definitely some opportunity to de-duplicate some common code. I decided to wait until we have more scenarios before doing this.

src/features/transaction-wizard/components/transaction-builder-form.tsx
```diff
@@ -0,0 +1,153 @@
...
+        transactionId,
+        transactionsGraphData,
+      })
+
+      toast.success('Transaction sent successfully')
+    },
+    [activeAddress, buildableTransaction, signer]
+  )
+
+  // NOTE: Number fields need the default value `'' as unknown as undefined` in order to be cleared.
```
PatrickDinh, line 80: Nice fix, I remember it now. This was the reason why we had form validation errors on the numbers before.

src/features/transaction-wizard/data/common.ts
```diff
@@ -0,0 +1,107 @@
...
+    .superRefine((fee, ctx) => {
+      if (!fee.setAutomatically && !fee.value) {
+        ctx.addIssue({
+          code: z.ZodIssueCode.custom,
+          message: 'Required',
+          path: ['value'],
+        })
+      }
+    }),
+}
```
PatrickDinh, line 60: I'm wondering if we should refactor this to
```
export const feeFieldSchema = z
  .object({
    setAutomatically: z.boolean(),
    value: numberSchema(z.number().min(0.001).optional()),
  })
  .superRefine((fee, ctx) => {
    if (!fee.setAutomatically && !fee.value) {
      ctx.addIssue({
        code: z.ZodIssueCode.custom,
        message: 'Required',
        path: ['value'],
      })
    }
  })
```
so that the consumer can decide the field name.
For example, 
```
const paymentSchema = zfd.formData({
  ...
  fee: feeFieldSchema,
  ....
})
```
vs 
```
const paymentSchema = zfd.formData({
  ...feeFieldSchema,
})
```

neilcampbell: We can adjust as we see more scenarios. My thought here is that fee will always be the same across all transactions, so It'd be strange to be using a different key per transaction.

src/features/transaction-wizard/components/transaction-builder-form.tsx
neilcampbell: It works, however it's a bit annoying having to do this. It also need to be coupled with the `numberSchema` function I created in order to work properly, so there are definitely a few traps/sharp edges.

src/features/forms/components/form-actions.tsx
```diff
@@ -1,11 +1,21 @@
-import { ReactNode } from 'react'
+import { ReactNode, useEffect, useState } from 'react'
 import { cn } from '@/features/common/utils'
 
 export interface FormActionsProps {
   className?: string
   children?: ReactNode
+  onInit?: () => void
```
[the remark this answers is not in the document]
neilcampbell: I remembered the other reason I put it here.
Using it with the key attribute, allows me to run init each time we selected a new transaction type to build without clearing the data in the form. Lifting it up to the form, wouldn't allow me to easily do this.

PR #248, opened 2024-09-02
chore: adding some tests for txn wizard payments

neilcampbell:


src/features/transaction-wizard/transaction-wizard-page.test.tsx
```diff
@@ -0,0 +1,253 @@
...
+            await user.click(sendButton)
+
+            const transactionId = await waitFor(
+              () => {
+                expect(component.queryByText('Required')).not.toBeInTheDocument()
+                const transactionIdComponent = getByDescriptionTerm(component.container, transactionIdLabel)
+                expect(transactionIdComponent).toBeDefined()
+                return transactionIdComponent.textContent!
+              },
+              { timeout: 10_000 }
```
PatrickDinh, line 125: I think this should be `5_000` because there is a max 5 seconds timeout for vitest test https://vitest.dev/api/#test.

PatrickDinh (approved):

src/features/transaction-wizard/transaction-wizard-page.test.tsx
neilcampbell: I upped it to 20 seconds in the vitest config as part of this PR.

Issue #187, opened 2024-08-09
[local-dev-experience-ui] Format ABI method information


PatrickDinh:
## Description

- When viewing a transaction of an application which has an arc4/arc32/arc56 app spec, the app call method, arguments and return value are formatted with prettier (or something similar) to make it shorter and are syntax highlighted similar to the JSON view to make easier to read.

## Acceptance Criteria

1. When viewing a transaction of an application which has an arc32 app spec, the app call method, arguments and return value are prettified and syntax highlighted.
1. When viewing a transaction of an application which has an arc4 app spec, the app call method, arguments and return value are prettified and syntax highlighted.
1. When viewing a transaction of an application which has an arc56 app spec, the app call method, arguments and return value are prettified and syntax highlighted.

neilcampbell:
We might be able to use this package https://www.npmjs.com/package/js-beautify

PR #254, opened 2024-09-06
Format ABI method information

negar-abbasi:
closes [#187](https://github.com/algorandfoundation/algokit-lora/issues/187)


src/features/abi-methods/components/decoded-abi-method-arguments.tsx
```diff
@@ -43,30 +47,46 @@ export function DecodedAbiMethodArguments({ method }: { method: AbiMethod }) {
...
-    <ul className="pl-4">
-      {components.map((component, index, arr) => (
-        <li key={index}>
-          {component}
-          {index < arr.length - 1 ? <span>{', '}</span> : null}
-        </li>
-      ))}
-    </ul>
-  )
+  const multiLine = argumentsRender.some((argument) => argument.multiLine) || sum(argumentsRender.map((argument) => argument.length)) > 20
```
PatrickDinh, line 66: I think this multiline logic should be moved to the mapper.

src/features/abi-methods/components/abi-value.tsx
```diff
@@ -14,7 +15,21 @@ export function AbiValue({ abiValue }: Props) {
     return <AbiArrayValue array={abiValue} />
   }
   if (abiValue.type === AbiType.String) {
-    return <span>{`"${abiValue.value}"`}</span>
+    return <span className="text-abi-string">{`"${abiValue.value}"`}</span>
+  }
+  if (abiValue.type === AbiType.Address) {
```
PatrickDinh, line 20: It looks like this doesn't support other reference types such as Account, Asset and Transaction

src/features/abi-methods/components/decoded-abi-method-arguments.tsx
```diff
@@ -5,16 +5,20 @@ import { TransactionLink } from '@/features/transactions/components/transaction-
...
     () =>
-      method.arguments.map((argument) => {
+      argumentsRender.map((argument) => {
         if (argument.type === AbiType.Transaction) {
           return (
             <>
-              <span>{argument.name}: </span>
-              <TransactionLink className="text-primary underline" transactionId={argument.value}>
+              <span className="text-abi-keys">{argument.name}: </span>
+              <TransactionLink className={'text-primary underline'} transactionId={argument.value}>
```
PatrickDinh, line 21:
```suggestion
              <TransactionLink className="text-primary underline" transactionId={argument.value}>
```
The same comment applied for other `className={'`

src/features/abi-methods/components/abi-value.tsx
PatrickDinh: my bad, they were handled in `DecodedAbiMethodArguments`, ignore this comment

src/features/abi-methods/components/abi-value.tsx
```diff
@@ -14,7 +15,21 @@ export function AbiValue({ abiValue }: Props) {
...
+    return (
+      <>
```
PatrickDinh, line 22: We don't need `<>` here

src/features/abi-methods/components/decoded-abi-method-arguments.tsx
```diff
@@ -5,15 +5,18 @@ import { TransactionLink } from '@/features/transactions/components/transaction-
...
+
   const components = useMemo(
     () =>
-      method.arguments.map((argument) => {
+      methodRepresentation.arguments.map((argument) => {
         if (argument.type === AbiType.Transaction) {
           return (
             <>
-              <span>{argument.name}: </span>
+              <span className="text-abi-keys">{argument.name}: </span>
```
PatrickDinh, line 19: It looks like this can be refactored so
```
<span className="text-abi-keys">{argument.name}: </span>
```
doesn't need to repeat

src/features/abi-methods/mappers/index.ts
```diff
@@ -20,3 +36,78 @@ export const jsonAsArc4AppSpec = (json: unknown): Arc4AppSpec => {
...
+export function getAbiMethodRepresentation(method: AbiMethod): AbiMethodRepresentation {
+  const argumentRepresentationList = method.arguments.map((argument) => asAbiMethodArgumentRepresentation(argument))
+  const multiLine =
+    argumentRepresentationList.some((argument) => argument.multiLine) ||
+    sum(argumentRepresentationList.map((arg) => arg.length)) > MAX_LENGTH
+  return {
+    name: method.name,
+    arguments: argumentRepresentationList,
+    multiLine: multiLine,
+    return: method.return,
```
PatrickDinh, line 73: we should map the return to `ReturnRepresentation` here as well

src/features/abi-methods/mappers/index.ts
```diff
@@ -2,6 +2,22 @@ import { Schema, Validator } from 'jsonschema'
...
+  AbiMethodRepresentation,
+  AbiTupleRepresentation,
+  AbiTupleValue,
+  AbiType,
+  AbiValue,
+  AbiValueRepresentation,
+} from '@/features/abi-methods/models'
+import { sum } from '@/utils/sum'
+
+const MAX_LENGTH = 20
```
PatrickDinh, line 20: I think `MAX_LINE_LENGTH` is a better name here.

src/features/abi-methods/components/decoded-abi-method-arguments.tsx
```diff
@@ -5,15 +5,18 @@ import { TransactionLink } from '@/features/transactions/components/transaction-
 import { AccountLink } from '@/features/accounts/components/account-link'
 import { ApplicationLink } from '@/features/applications/components/application-link'
 import { AssetIdLink } from '@/features/assets/components/asset-link'
+import { getAbiMethodRepresentation } from '@/features/abi-methods/mappers'
 
 export function DecodedAbiMethodArguments({ method }: { method: AbiMethod }) {
+  const methodRepresentation = useMemo(() => getAbiMethodRepresentation(method), [method])
```
PatrickDinh, line 11: This mapping logic in `getAbiMethodRepresentation` is quite expensive because we have to recursively build the arguments. Therefore, this component and `DecodedAbiMethodReturnValue` should be refactored to received `AbiMethodArgumentRepresentation` and `AbiMethodReturnRepresentation` respectively so that we don't need to run the mapping logic twice.

src/features/abi-methods/components/abi-value.tsx
```diff
@@ -14,7 +15,19 @@ export function AbiValue({ abiValue }: Props) {
     return <AbiArrayValue array={abiValue} />
   }
   if (abiValue.type === AbiType.String) {
-    return <span>{`"${abiValue.value}"`}</span>
+    return <span className="text-abi-string">{`"${abiValue.value}"`}</span>
+  }
+  if (abiValue.type === AbiType.Number) {
+    return <span className="text-abi-number">{`${abiValue.value}`}</span>
```
PatrickDinh, line 21:
```suggestion
    return <span className="text-abi-number">{abiValue.value}</span>
```
same for the bool value

src/features/abi-methods/components/abi-value.tsx
```diff
@@ -14,7 +15,19 @@ export function AbiValue({ abiValue }: Props) {
...
+    return <span className="text-abi-number">{abiValue.value}</span>
+  }
+  if (abiValue.type === AbiType.Boolean) {
+    return <span className="text-abi-bool">{abiValue.value}</span>
+  }
+  if (abiValue.type === AbiType.Address) {
+    return (
+      <AccountLink className="text-primary underline" address={abiValue.value}>
+        {abiValue.value}
+      </AccountLink>
```
PatrickDinh:
```suggestion
      <AccountLink address={abiValue.value} />
```

src/features/abi-methods/components/decoded-abi-method-arguments.tsx
```diff
@@ -1,72 +1,82 @@
...
+}: {
+  arguments: AbiMethodArgumentRepresentation[]
+  multiLine: boolean
+}) {
+  const renderArgumentValue = (argument: AbiMethodArgumentRepresentation) => {
+    if (argument.type === AbiType.Transaction) {
+      return (
+        <TransactionLink className="text-primary underline" transactionId={argument.value}>
+          {argument.value}
+        </TransactionLink>
```
PatrickDinh, line 21:
```suggestion
        <TransactionLink transactionId={argument.value} />
```
same for other links.

src/features/abi-methods/components/decoded-abi-method-arguments.tsx
```diff
@@ -1,72 +1,82 @@
...
 
-export function DecodedAbiMethodArguments({ method }: { method: AbiMethod }) {
+export function DecodedAbiMethodArguments({
+  arguments: argumentsProp,
+  multiLine,
+}: {
+  arguments: AbiMethodArgumentRepresentation[]
+  multiLine: boolean
+}) {
+  const renderArgumentValue = (argument: AbiMethodArgumentRepresentation) => {
```
PatrickDinh, line 16: This method needs to be wrapped inside a `useCallback`, otherwise it may cause pref issue because it's calculated on every render.

src/features/abi-methods/components/decoded-abi-method.tsx
```diff
@@ -1,21 +1,23 @@
...
 import { DecodedAbiMethodArguments } from '@/features/abi-methods/components/decoded-abi-method-arguments'
 import { AbiMethod } from '@/features/abi-methods/models'
+import { getAbiMethodRepresentation } from '../mappers'
 
 type Props = {
   abiMethod: AbiMethod
 }
 
 export function DecodedAbiMethod({ abiMethod }: Props) {
+  const abiMethodRepresentation = getAbiMethodRepresentation(abiMethod)
```
PatrickDinh, line 11: this needs to be inside a `useMemo`

src/features/abi-methods/mappers/index.ts
```diff
@@ -20,3 +37,82 @@ export const jsonAsArc4AppSpec = (json: unknown): Arc4AppSpec => {
...
+    abiMethodArgument.type === AbiType.Transaction ||
+    abiMethodArgument.type === AbiType.Asset
+  ) {
+    return {
+      ...abiMethodArgument,
+      multiLine: false,
+      length: `${abiMethodArgument.value}`.length,
+    }
+  }
+  const resolvedArgument = asAbiValueRepresentation(abiMethodArgument)
```
PatrickDinh, line 55: rename this var to `abiValueRepresentation` to reflect the type

src/features/abi-methods/mappers/index.ts
```diff
@@ -20,3 +37,82 @@ export const jsonAsArc4AppSpec = (json: unknown): Arc4AppSpec => {
...
+  const multiLine = sum([resolvedArgument.length]) > MAX_LINE_LENGTH
+
+  return {
+    ...abiMethodArgument,
+    ...resolvedArgument,
+    multiLine: multiLine,
+  }
```
PatrickDinh, line 62: maybe I am missing some context, but I have some questions about this section:
- sum of an array with just 1 element isn't needed.
- the returned value of `asAbiValueRepresentation` should indicate the flag `multiLine`, why do you need to re-calculate it here?

src/features/abi-methods/mappers/index.ts
```diff
@@ -20,3 +37,82 @@ export const jsonAsArc4AppSpec = (json: unknown): Arc4AppSpec => {
...
+  return {
+    ...abiValue,
+    multiLine: false,
+    length: `${abiValue.value}`.length,
+  }
+}
+
+const asAbiTupleRepresentation = (abiTuple: AbiTupleValue): AbiTupleRepresentation => {
+  const valueRepresentation = abiTuple.values.map((value) => asAbiValueRepresentation(value))
```
PatrickDinh, line 97: maybe rename to `valueRepresentations`

src/features/abi-methods/mappers/index.ts
```diff
@@ -20,3 +37,82 @@ export const jsonAsArc4AppSpec = (json: unknown): Arc4AppSpec => {
...
+  return {
+    type: AbiType.Tuple,
+    values: valueRepresentation,
+    multiLine: multiLine,
+    length: length,
+  }
+}
+
+const asAbiArrayRepresentation = (abiArray: AbiArrayValue): AbiArrayRepresentation => {
+  const valueRepresentation = abiArray.values.map((value) => asAbiValueRepresentation(value))
```
PatrickDinh, line 109: maybe rename to `valueRepresentations`

PatrickDinh (approved):
LGTM

PR #259, opened 2024-09-17
feat: confirm resource packing

PatrickDinh:


src/features/forms/components/transaction-form-item.tsx
```diff
@@ -133,67 +134,60 @@ function TransactionBuilder<TSchema extends Record<string, unknown>>({
...
       {savedValues && (
         <>
           <DescriptionList items={thign} />
-          <Button onClick={buildTransaction}>Edit</Button>
+          <Button type="button" onClick={openDialog}>
+            Edit
+          </Button>
         </>
       )}
-      <FormItem {...props} field={field} disabled={disabled}>
```
PatrickDinh, line 186: @neilcampbell I removed this hidden input because the parent form failed to reset with it being here. Do we need this input?

neilcampbell: Yep, the transaction information is JSON serialised into that input and that's the transaction is materialised.
This is unless you've changed the logic to do it differently?

neilcampbell (approved):

PR #261, opened 2024-09-24
fix: validating values in application global state

### LINKED ISSUE #255: App Global States not decoding correctly
The same app may have states encoded in different formats, lora seems to detect the first format and apply it to all subsequent ones

example
[allo](https://allo.info/application/765919289) vs [lora](https://lora.algokit.io/mainnet/application/765919289)

negar-abbasi:
closes #255
Screenshot 2024-09-24 at 12 22 35 PM

### DIFF
--- a/src/features/applications/mappers/index.ts
+++ b/src/features/applications/mappers/index.ts
@@ -15,7 +15,6 @@ import { asJson } from '@/utils/as-json'
 import { Arc32AppSpec, Arc4AppSpec } from '@/features/app-interfaces/data/types'
 import algosdk from 'algosdk'
 import { isArc32AppSpec } from '@/features/common/utils'
-import { isAddress } from '@/utils/is-address'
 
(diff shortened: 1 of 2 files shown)


src/features/applications/mappers/index.ts
```diff
@@ -91,14 +92,19 @@ const getGlobalStateValue = (tealValue: modelsv2.TealValue): ApplicationGlobalSt
...
-    } else {
-      return buf.toString('base64')
+    const encodedAddress = encodeAddress(new Uint8Array(buf))
+    if (isAddress(encodedAddress)) {
+      return encodedAddress
+    }
+  }
+  if (isUtf8(buf)) {
+    const utf8 = buf.toString('utf8')
+    const printableUtf8Regex = /^[ -~]+$/
```
neilcampbell, line 102: I think this might be ASCII chars rather than utf8.

neilcampbell (approved):

PR #306, opened 2024-10-30
AK-14 - Added page titles

lempira:
I looks through all of the URL routes and added a hook that creates the title based on the params of page. Here are the pages and examples of titles. I still have to write a test for the hook but wanted get feedback on approach and format before I write the test.

Pages: 
"Index": "/",
  Example: Lora Explore - mainnet
  
  "Explore": "/:networkId",
  Example: Lora Explore - mainnet
  
  "Explore_Transaction_ById": "/:networkId/transaction/:transactionId",
  Example: Lora - TxnId:CFNKY4JV5BMIRSP2SMZE634DYB22FALLURI72U6UMWZNKEBOYEMQ - mainnet
  
  "Explore_Transaction_ById_Inner_ById": "/:networkId/transaction/:transactionId/inner//*",
  Example: Lora - TxnId:Q36W3E5GBCPZYGEVFQGTQB4KX7QTFYV3H4UYNRYS4AFXEPILHQXQ, Inner:1 - mainnet
  
  "Explore_Block_ByRound": "/:networkId/block/:round",
  Example: Lora - Block:43904244 - mainnet
  
  "Explore_Block_ByRound_Group_ById": "/:networkId/block/:round/group/:groupId",
  Example: Lora - Block:43904244 - Group:iUE01XrzcJZ+ENIwPyjPmtaVUF5OKy8vYYOwGbdKliQ= - mainnet
  
  "Explore_Account_ByAddress": "/:networkId/account/:address",
  Example: Lora - Addr:W2IZ3EHDRW2IQNPC33CI2CXSLMFCFICVKQVWIYLJWXCTD765RW47ONNCEY - mainnet
  
  "Explore_Asset_ById": "/:networkId/asset/:assetId",
  Example: Lora  - AssetId:1014 - localnet
  
  "Explore_Application_ById": "/:networkId/application/:applicationId",
  Example: Lora  - AppId:1019 - localnet
  
  "Explore_Tx": "/:networkId/tx/*",
  Example: Lora - TxnId:CFNKY4JV5BMIRSP2SMZE634DYB22FALLURI72U6UMWZNKEBOYEMQ - mainnet
  
  "Explore_Txn": "/:networkId/txn/*",
  Example: Lora - TxnId:CFNKY4JV5BMIRSP2SMZE634DYB22FALLURI72U6UMWZNKEBOYEMQ - mainnet
  
  "AppLab": "/app-lab",
  Example: Lora App Lab
  
  "AppLab_Create": "/app-lab/create",
  Example: Lora Create App Interface
  
  "Settings": "/settings",
  Example: Lora Settings
  
  "Fund": "/fund",
  Example: Lora Fund
  
  "FundAuthCallback": "/fund/auth-callback",
  Example: NA
  
  "TransactionWizard": "/transaction-wizard",
  Example: Lora Transaction Wizard
  
  "TxWizard": "/tx-wizard",
  Example: Lora Transaction Wizard
  
  "TxnWizard": "/txn-wizard"
  Example: Lora Transaction Wizard

### DIFF
--- a/src/features/explore/pages/explore-page.test.tsx
+++ b/src/features/explore/pages/explore-page.test.tsx
@@ -1,7 +1,7 @@
 import { executeComponentTest } from '@/tests/test-component'
 import { getAllByRole, getByRole, queryAllByRole, render, waitFor } from '@/tests/testing-library'
 import { Atom, createStore } from 'jotai'
-import { describe, expect, it } from 'vitest'
+import { describe, expect, it, vi } from 'vitest'
 import { ExplorePage } from './explore-page'
 import { latestBlocksTitle } from '@/features/blocks/components/latest-blocks'
 import { latestTransactionsTitle } from '@/features/transactions/components/latest-transactions'
@@ -16,12 +16,14 @@ import { randomNumberBetween } from '@makerx/ts-dossier'
 import { ellipseId } from '@/utils/ellipse-id'
 import { ellipseAddress } from '@/utils/ellipse-address'
 import { createReadOnlyAtomAndTimestamp, createTimestamp } from '@/features/common/data'
+import { useParams } from 'react-router-dom'
 
 describe('explore-page', () => {
   describe('when no blocks are available', () => {
     const myStore = createStore()
 
-    it('no latest blocks are displayed', () => {
+    it.only('no latest blocks are displayed', () => {
+      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
@@ -36,6 +38,7 @@ describe('explore-page', () => {
     })
 
     it('no latest transactions are displayed', () => {
+      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
@@ -64,6 +67,7 @@ describe('explore-page', () => {
     myStore.set(syncedRoundAtom, block.round)
 
     it('the processed blocks are displayed', () => {
+      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
@@ -83,6 +87,7 @@ describe('explore-page', () => {
     })
 
     it('the available transactions are displayed', () => {
+      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
@@ -130,6 +135,7 @@ describe('explore-page', () => {
       myStore.set(transactionResultsAtom, data.transactions)
       myStore.set(blockResultsAtom, data.blocks)
       myStore.set(syncedRoundAtom, data.syncedRound)
+      vi.mocked(useParams).mockReturnValue({ round: '1234' })
 
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
@@ -156,6 +162,7 @@ describe('explore-page', () => {
       myStore.set(blockResultsAtom, data.blocks)
       myStore.set(syncedRoundAtom, data.syncedRound)
 
+      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
--- a/src/features/explore/pages/explore-page.test.tsx
+++ b/src/features/explore/pages/explore-page.test.tsx
@@ -59,7 +59,7 @@ describe('explore-page', () => {
     const block = blockResultMother.blockWithTransactions(transactionResults).withTimestamp(1719284618).build()
     const myStore = createStore()
     myStore.set(blockResultsAtom, new Map([[block.round, createReadOnlyAtomAndTimestamp(block)]]))
-    myStore.set(transactionResultsAtom, new Map(transactionResults.map((x) => [x.id, createReadOnlyAtomAndTimestamp(x)])))
+    myStore.set(transactionResultsAtom, new Map(transactionResults.map((t) => [t.id, createReadOnlyAtomAndTimestamp(t)])))
     myStore.set(
       latestTransactionIdsAtom,
       transactionResults.map((t) => [t.id, createTimestamp()] as const)
--- a/src/features/explore/pages/explore-page.test.tsx
+++ b/src/features/explore/pages/explore-page.test.tsx
@@ -1,7 +1,7 @@
 import { executeComponentTest } from '@/tests/test-component'
 import { getAllByRole, getByRole, queryAllByRole, render, waitFor } from '@/tests/testing-library'
 import { Atom, createStore } from 'jotai'
-import { describe, expect, it, vi } from 'vitest'
+import { describe, expect, it } from 'vitest'
 import { ExplorePage } from './explore-page'
 import { latestBlocksTitle } from '@/features/blocks/components/latest-blocks'
 import { latestTransactionsTitle } from '@/features/transactions/components/latest-transactions'
@@ -16,14 +16,12 @@ import { randomNumberBetween } from '@makerx/ts-dossier'
 import { ellipseId } from '@/utils/ellipse-id'
 import { ellipseAddress } from '@/utils/ellipse-address'
 import { createReadOnlyAtomAndTimestamp, createTimestamp } from '@/features/common/data'
-import { useParams } from 'react-router-dom'
 
 describe('explore-page', () => {
   describe('when no blocks are available', () => {
     const myStore = createStore()
 
     it('no latest blocks are displayed', () => {
-      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
@@ -38,7 +36,6 @@ describe('explore-page', () => {
     })
 
     it('no latest transactions are displayed', () => {
-      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
@@ -67,7 +64,6 @@ describe('explore-page', () => {
     myStore.set(syncedRoundAtom, block.round)
 
     it('the processed blocks are displayed', () => {
-      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
@@ -87,7 +83,6 @@ describe('explore-page', () => {
     })
 
     it('the available transactions are displayed', () => {
-      vi.mocked(useParams).mockReturnValue({ round: '1234' })
       return executeComponentTest(
         () => render(<ExplorePage />, undefined, myStore),
         async (component) => {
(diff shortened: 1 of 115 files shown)


src/utils/use-title.ts
```diff
@@ -0,0 +1,47 @@
+import { useEffect } from 'react'
+import { useParams } from 'react-router-dom'
+
+const setTitle = (title: string) => {
+  document.title = title
+}
+
+export const useTitle = (customString?: string) => {
```
neilcampbell, line 8:
```suggestion
export const useTitle = (pagePrefix?: string) => {
```
My thought process here is that anything after the `Lora - ` is the page specific portion of the title, so it's a page prefix, even though it's not the title prefix.

src/utils/use-title.ts
```diff
@@ -0,0 +1,47 @@
...
+  const urlParams = useParams()
+  useEffect(() => {
+    const currentTitle = document.title
+    let pageTitle = `Lora`
```
neilcampbell, line 12: I think it would be good for all pages to start with `Lora - ${pageTitle}`
So:
`Lora - Transaction Wizard`
`Lora - TxnId:CFNKY4JV5BMIRSP2SMZE634DYB22FALLURI72U6UMWZNKEBOYEMQ - mainnet`

Probably the easiest way to do this is accumulate in an array, then `.join(' - ')` to build the page title. Something like below:

```
  const params = new Array<string>()

  if (urlParams.transactionId) {
    params.push(`TxnId:${urlParams.transactionId}`)
  }
  // ...

  const title = `Lora${pagePrefix ? ` ${pagePrefix}` : ''} - ${params.join(' - ')}`
```

neilcampbell (commented):
The approach looks good to me 👍

src/utils/use-title.ts
tasosbit: I have a small preference for omitting the `Id` part in transactions / assets / apps

So: `Txn:${id}` `App:${id}` `Asset:${id}` 

What do you think?

src/utils/use-title.ts
```diff
@@ -0,0 +1,47 @@
...
+      pageTitle += `, Inner:${urlParams['*']}`
+    }
+    if (urlParams.round) {
+      pageTitle += ` - Block:${urlParams.round}`
+    }
+    if (urlParams?.groupId) {
+      pageTitle += ` - Group:${urlParams.groupId}`
+    }
+    if (urlParams?.address) {
+      pageTitle += ` - Addr:${urlParams.address}`
```
tasosbit, line 29: The verbiage we use in these pages is `Account` (e.g. check <H1> there), so the document title should probably be `Account` or `Acct` instead of `Addr` to match that

tasosbit: And if we omit the `-` separators, browser tabs would show more of the title before truncating:

`Lora Acct ACSASDSADASD Mainnet` vs `Lora - Acct:ACSASDSADASD - Mainnet`

src/utils/use-title.ts
tasosbit: And if we omit the `-` separators, browser tabs would show more of the title before truncating:

`Lora Txn ACSASDSADASD Mainnet` vs `Lora - Txn:ACSASDSADASD - Mainnet`

May not make much difference though, so this is quite subjective

neilcampbell: I'm happy with both of those suggestions.

lempira: Me too. I'll make these updates today.

src/utils/use-title.ts
lempira: 👍 Will change to use Acct

lempira:
I changed the page title format according to the suggested format. I also used @neilcampbell suggestion of joining the params into a string.

For the use-title test, I use a createMemoryBrowser of react-router-dom instead of mocking the useParams. I think this allows us to test more declaratively and pass in that URL instead of imperatively passing in the params. 

There are many places that the params are mocked thru `vi.mocked(useParams).mockImplementation`. Perhaps we can create a helper wrapper that uses createMemoryBrowser and pass in the URL and not have to mock useParams?

Another suggestion: can we consider adding [eslint-plugin-vitest](https://github.com/vitest-dev/eslint-plugin-vitest) to the testing conf. I had some focused tests that I forgot to unfocus, and they were only caught in the pipeline. The eslint plugin could have caught that, along with whatever other rules we want to enforce, on lint.

neilcampbell:
> There are many places that the params are mocked thru vi.mocked(useParams).mockImplementation. Perhaps we can create a helper wrapper that uses createMemoryBrowser and pass in the URL and not have to mock useParams?

> Another suggestion: can we consider adding [eslint-plugin-vitest](https://github.com/vitest-dev/eslint-plugin-vitest) to the testing conf. I had some focused tests that I forgot to unfocus, and they were only caught in the pipeline. The eslint plugin could have caught that, along with whatever other rules we want to enforce, on lint.

Both of these sound good.

src/utils/use-title.ts
```diff
@@ -0,0 +1,47 @@
...
+}
+
+export const useTitle = (pagePrefix?: string) => {
+  const urlParams = useParams()
+
+  useEffect(() => {
+    const currentTitle = document.title
+    const pageTitleParams: string[] = []
+    if (urlParams.transactionId) {
+      pageTitleParams.push(`Txn:${urlParams.transactionId.slice(0, TRUNCATE_LENGTH)}`)
```
neilcampbell, line 17:
```suggestion
      pageTitleParams.push(`Txn:${ellipseId(urlParams.transactionId)}`)
```
This will inherit the same convention as we use elsewhere.

src/utils/use-title.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (urlParams.transactionId && urlParams['*']) {
+      pageTitleParams.push(`Inner:${urlParams['*']}`)
+    }
+    if (urlParams.round) {
+      pageTitleParams.push(`Block:${urlParams.round}`)
+    }
+    if (urlParams?.groupId) {
+      pageTitleParams.push(`Group:${urlParams.groupId.slice(0, TRUNCATE_LENGTH)}`)
```
neilcampbell, line 26:
```suggestion
      pageTitleParams.push(`Group:${ellipseId(urlParams.groupId)}`)
```

src/utils/use-title.ts
```diff
@@ -0,0 +1,47 @@
...
+    }
+    if (urlParams?.address) {
+      pageTitleParams.push(`Acct:${urlParams.address.slice(0, TRUNCATE_LENGTH)}`)
```
neilcampbell, line 29:
```suggestion
      pageTitleParams.push(`Acct:${ellipseAddress(urlParams.address)}`)
```

src/utils/use-title.ts
```diff
@@ -0,0 +1,47 @@
+import { useEffect } from 'react'
+import { useParams } from 'react-router-dom'
+
+const TRUNCATE_LENGTH = 10
```
neilcampbell, line 4:
```suggestion
```

src/features/explore/pages/explore-page.test.tsx
```diff
@@ -16,12 +16,14 @@ import { randomNumberBetween } from '@makerx/ts-dossier'
...
 import { ellipseAddress } from '@/utils/ellipse-address'
 import { createReadOnlyAtomAndTimestamp, createTimestamp } from '@/features/common/data'
+import { useParams } from 'react-router-dom'
 
 describe('explore-page', () => {
   describe('when no blocks are available', () => {
     const myStore = createStore()
 
     it('no latest blocks are displayed', () => {
+      vi.mocked(useParams).mockReturnValue({ round: '1234' })
```
neilcampbell, line 26:
```suggestion
```
AFAIK round isn't used as a param here. I also pushed a small change (as part of merging main into the branch) which removes the need for `vi.mocked(useParams).mockReturnValue({})` in the tests that don't use params.

src/utils/use-title.ts
lempira: 👍 made the change

src/utils/use-title.ts
lempira: 👍 made the change

src/utils/use-title.ts
lempira: 👍 made the change

src/utils/use-title.ts
lempira: 👍 made the change

lempira:
> > There are many places that the params are mocked thru vi.mocked(useParams).mockImplementation. Perhaps we can create a helper wrapper that uses createMemoryBrowser and pass in the URL and not have to mock useParams?
> 
> > Another suggestion: can we consider adding [eslint-plugin-vitest](https://github.com/vitest-dev/eslint-plugin-vitest) to the testing conf. I had some focused tests that I forgot to unfocus, and they were only caught in the pipeline. The eslint plugin could have caught that, along with whatever other rules we want to enforce, on lint.
> 
> Both of these sound good. @lempira Did you want to do these as part of this PR or handle separately?

I was thinking taking these items to a separate PR to keep this one smaller.

src/tests/utils/use-title.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    routePath: '/:networkId',
+    url: '/mainnet',
+    pagePrefix: 'Explore',
+    expectedText: 'Lora Explore mainnet',
+  },
+  {
+    routePathName: 'Explore_Transaction_ById',
+    routePath: '/:networkId/transaction/:transactionId',
+    url: '/mainnet/transaction/CFNKY4JV5BMIRSP2SMZE634DYB22FALLURI72U6UMWZNKEBOYEMQ',
+    expectedText: 'Lora Txn:CFNK…YEMQ mainnet',
```
neilcampbell, line 30:
```suggestion
    expectedText: 'Lora Txn:CFNKY4J… mainnet',
```

src/utils/use-title.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export const useTitle = (pagePrefix?: string) => {
+  const urlParams = useParams()
+
+  useEffect(() => {
+    const currentTitle = document.title
+    const pageTitleParams: string[] = []
+    if (urlParams.transactionId) {
+      pageTitleParams.push(`Txn:${ellipseAddress(urlParams.transactionId)}`)
```
neilcampbell, line 16:
```suggestion
      pageTitleParams.push(`Txn:${ellipseId(urlParams.transactionId)}`)
```

src/utils/use-title.ts
```diff
@@ -0,0 +1,46 @@
...
+    }
+    if (urlParams.transactionId && urlParams['*']) {
+      pageTitleParams.push(`Inner:${urlParams['*']}`)
+    }
+    if (urlParams.round) {
+      pageTitleParams.push(`Block:${urlParams.round}`)
+    }
+    if (urlParams?.groupId) {
+      pageTitleParams.push(`Group:${ellipseAddress(urlParams.groupId)}`)
```
neilcampbell, line 25:
```suggestion
      pageTitleParams.push(`Group:${ellipseId(urlParams.groupId)}`)
```

src/tests/utils/use-title.test.tsx
```diff
@@ -0,0 +1,92 @@
...
+    routePathName: 'Explore_Block_ByRound',
+    routePath: '/:networkId/block/:round',
+    url: '/mainnet/block/43904244',
+    expectedText: 'Lora Block:43904244 mainnet',
+  },
+  {
+    routePathName: 'Explore_Block_ByRound_Group_ById',
+    routePath: '/:networkId/block/:round/group/:groupId',
+    url: '/mainnet/block/43904244/group/iUE01XrzcJZ+ENIwPyjPmtaVUF5OKy8vYYOwGbdKliQ=',
+    expectedText: 'Lora Block:43904244 Group:iUE0…liQ= mainnet',
```
neilcampbell, line 48:
```suggestion
    expectedText: 'Lora Block:43904244 Group:iUE01Xr… mainnet',
```

neilcampbell (approved):

PR #307, opened 2024-10-30
chore: heavier header fonts

### LINKED ISSUE #283: Improve text hierarchy
The application currently uses identical font, font size, and font weight for all text elements, including headers and value fields. This uniform styling is applied across the entire page, regardless of the text's purpose or importance.

## Issue
The lack of visual hierarchy in the text makes it difficult for users to:
1. Quickly scan and understand the structure of the page
2. Distinguish between different types of information (e.g., headers vs. content)
3. Identify the most important information at a glance

This is most obvious on the "header" tables above the main tables on pages like block and account, where the left header values look exactly the same as value fields on the right. It would be enough to simply increase their font weight to make them different.

negar-abbasi:
closes #283

Screenshot 2024-10-30 at 1 36 23 PM

### DIFF
--- a/src/features/common/components/description-list.tsx
+++ b/src/features/common/components/description-list.tsx
@@ -15,7 +15,7 @@ export function DescriptionList({ items, dtClassName }: Props) {
     <div className={cn('grid grid-cols-[minmax(min-content,auto)_1fr] gap-x-4 gap-y-1.5')}>
       {items.map((item, index) => (
         <dl key={index} className={cn('grid grid-cols-subgrid col-span-2')}>
-          <dt className={cn(dtClassName)}>{item.dt}</dt>
+          <dt className={cn(dtClassName, 'font-medium')}>{item.dt}</dt>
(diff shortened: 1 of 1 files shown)


src/features/common/components/description-list.tsx
```diff
@@ -15,7 +15,7 @@ export function DescriptionList({ items, dtClassName }: Props) {
     <div className={cn('grid grid-cols-[minmax(min-content,auto)_1fr] gap-x-4 gap-y-1.5')}>
       {items.map((item, index) => (
         <dl key={index} className={cn('grid grid-cols-subgrid col-span-2')}>
-          <dt className={cn(dtClassName)}>{item.dt}</dt>
+          <dt className={cn(dtClassName, 'font-medium')}>{item.dt}</dt>
```
neilcampbell, line 18:
```suggestion
          <dt className={cn('font-medium', dtClassName)}>{item.dt}</dt>
```

Ensures the `dtClassName` classes will take precedence.

neilcampbell (approved):

PR #329, opened 2024-11-08
test for preview

lempira:


src/features/accounts/pages/account-page.tsx
```diff
@@ -18,7 +18,9 @@ export const accountFailedToLoadMessage = 'Account failed to load'
 
 const transformError = (e: Error) => {
   if (is404(e)) {
-    return new Error(accountInvalidAddressMessage)
+    const error = new Error(accountInvalidAddressMessage) as StatusError
+    error.status = 404
+    return error
```
lempira, line 23: I had to add this to each of the transformErrors in each of pages because it strips out an identification that the message is a 404. I used this info the error-page to display the custom error message

src/features/network/pages/network-page.tsx
```diff
@@ -11,28 +11,26 @@ type Props = {
...
+    if (currentNetworkId === selectedNetwork) {
+      return
     }
-    // eslint-disable-next-line react-hooks/exhaustive-deps
-  }, [networkId, selectedNetwork])
+
+    setSelectedNetwork(currentNetworkId)
+    const newUrl = pathname.replace(selectedNetwork, currentNetworkId)
+    navigate(newUrl, { replace: true })
+  }, [currentNetworkId, navigate, pathname, selectedNetwork, setSelectedNetwork])
```
lempira, line 33: The new behavior of the useEffect is to align the currentNetworkId (i.e. from the URL) and the selectedNetworkId (i.e. from the stored value) if there is a mismatch

src/features/network/pages/network-page.tsx
```diff
@@ -11,28 +11,26 @@ type Props = {
...
 
   useEffect(() => {
-    if (networkId === wildcardNetworkRoute) {
-      // Handle the wildcard network route
-      navigate(pathname.replace(wildcardNetworkRoute, selectedNetwork) + search + hash, { replace: true })
-    } else if (networkId !== selectedNetwork) {
-      // When a user changes the network, their history will contain routes for the previous network.
-      // This handles navigating the user to the explore page for the selected network, so they don't get 404s when navigating through history.
-      navigate(Urls.Network.Explore.build({ networkId: selectedNetwork }), { replace: true })
+    if (currentNetworkId === selectedNetwork) {
```
neilcampbell, line 26: Previously we had a behaviour which allowed users to hit a route with and `_` as the network and it'd select the last selected network. This no longer works.

src/features/network/components/network-select.tsx
```diff
@@ -3,37 +3,43 @@ import { Label } from '@/features/common/components/label'
...
-      <Label htmlFor="network" className={cn('ml-0.5 mb-2')}>
-        Active network
-      </Label>
+    <div className={cn('flex flex-col')}>
+      {showLabel && (
+        <Label htmlFor="network" className={cn('ml-0.5 mb-2')}>
+          Active network
+        </Label>
+      )}
       <Select onValueChange={handleNetworkChange} value={selectedNetwork}>
```
neilcampbell, line 38: The focus ring looks like it's getting cut off on the left side. See pic.

Screenshot 2024-11-22 at 11 37 50 PM

src/features/network/components/network-select.tsx
```diff
@@ -3,37 +3,43 @@ import { Label } from '@/features/common/components/label'
...
   const navigate = useNavigate()
   const location = useLocation()
 
   const handleNetworkChange = useCallback(
-    async (value: string) => {
-      if (location.pathname !== settingsUrl) {
-        navigate(settingsUrl)
+    (newNetworkId: string) => {
+      const currentPath = location.pathname
+      setSelectedNetwork(newNetworkId)
```
neilcampbell, line 22:
```suggestion
      setSelectedNetwork(newNetworkId)
```
I don't believe this is needed. With this here, we end up calling setSelectedNetwork twice.

src/features/network/components/network-configs-table.tsx
```diff
@@ -75,10 +78,20 @@ const tableColumns: ColumnDef<NetworkConfigWithId>[] = [
...
+    header: '',
+    meta: { className: `w-[${ICON_BUTTON_SIZE}px]` },
+    accessorFn: (item) => item,
+    cell: (cell) => {
+      const networkConfig = cell.getValue<NetworkConfigWithId>()
+      return <ActivateButton networkConfig={networkConfig} />
+    },
+  },
   {
     id: 'edit',
```
neilcampbell, line 92: I think it may be better to bring these buttons into a single "Actions" (no need for a visible header) column and reduce the padding a little between the buttons. The icon also feels a bit big for the button. The app interface page has an edit and delete icon button, when you have. The styling and padding should match this.

Screenshot 2024-11-22 at 10 41 55 PM

Also when reducing the screen width, the buttons appear as below.
Screenshot 2024-11-22 at 10 42 37 PM

src/features/network/components/network-select.tsx
neilcampbell: I discovered why it's here. Without it you can select the network from the settings page or a page without network in the url. I wonder if there is a way to not call `setSelectedNetwork` twice?

src/features/app-interfaces/pages/edit-app-interface-page.tsx
```diff
@@ -12,7 +12,15 @@ import { useTitle } from '@/utils/use-title'
 
 const transformError = (e: Error) => {
   if (is404(e)) {
-    return new Error(appInterfaceNotFoundMessage)
+    const error = new Error(appInterfaceNotFoundMessage) as StatusError
+    error.status = 404
+    return error
+  }
+  // This is needed because the App interface not found doesn't return a 404 status code
+  if (e.message.includes('App interface not found')) {
```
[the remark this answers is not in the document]
lempira: I am not sure what you mean here? I wasn't seeing that the 404 was being triggered when the app interface was not found. What would we be extracting? The reason for that text is that that was being return by the server in the error
image

src/features/common/pages/error-page.tsx
```diff
@@ -1,20 +1,31 @@
...
     <>
       <PageTitle title={title ?? 'Error'} />
       <div>
         <p>Error: {error.message}</p>
       </div>
+      {is404(error) && redirectUrl && (
+        <div className="mt-4">
+          <p>Are you sure you have the correct network selected?</p>
+          <Link to={redirectUrl.build({ networkId: selectedNetwork })} className="text-primary hover:underline">
+            Go to main feature page
```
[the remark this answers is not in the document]
lempira: Made the change

src/features/layout/components/header.tsx
```diff
@@ -24,9 +25,10 @@ export function Header({ className }: Props) {
         <Search />
       </div>
       <div className={cn('flex items-center gap-4 ml-auto overflow-hidden')}>
-        <TemplatedNavLink urlTemplate={Urls.Settings} className="ml-2 truncate">
+        <NetworkSelect showLabel={false} />
+        {/* <TemplatedNavLink urlTemplate={Urls.Settings} className="ml-2 truncate">
```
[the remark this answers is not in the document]
lempira: Yup. Deleted

src/features/network/components/network-select.tsx
```diff
@@ -3,37 +3,43 @@ import { Label } from '@/features/common/components/label'
...
-import { Urls } from '@/routes/urls'
+import { useLocation, useNavigate, useParams } from 'react-router-dom'
 
-const settingsUrl = Urls.Settings.build({})
+type NetworkSelectProps = {
+  showLabel?: boolean
+}
 
-export function NetworkSelect() {
+export function NetworkSelect({ showLabel = true }: NetworkSelectProps) {
```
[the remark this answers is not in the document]
lempira: It seem to be used in the token-prompt-dialog
image

src/features/network/pages/network-page.tsx
lempira: I added this functionality back in Out of curiosity, why was this added?

src/features/network/components/network-select.tsx
lempira: fixed 
image

src/features/network/components/network-select.tsx
lempira: Perhaps I am not understanding. Where is the second time the `setSelectedNetwork` is being called? When you change the drop-down, the `setSelectedNetwork` is called once, and the URL is replaced with the new network, which ensures the URL that the selected network in state is the same.

src/features/network/components/network-configs-table.tsx
lempira: Good call. Made the change. I also noticed that the cells are hidden on small screen sizes but the table headers are not as shown in the screenshot above. I applied the same `meta.className` to the headers in the DataTable as well so they get hidden on small screen sizes. Are there cases where you want the `meta.className` only to applied to the cells and not the header?

image

src/features/app-interfaces/pages/edit-app-interface-page.tsx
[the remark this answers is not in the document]
neilcampbell: I noticed we have 2 different messages for the App inteface not found state. "App interface not found" and "Application Interface not found". We should consolidate this into one. Additionally we should extract this into a constant and use that const when doing the conditional and where the error is throw to ensure that if the message changes the conditional also changes.

src/features/network/components/network-select.tsx
neilcampbell: In the effect on the network-page. If you put a console.log just before both setSelectedNetwork calls, they both fire when changing the network.

src/features/network/pages/network-page.tsx
neilcampbell: I honestly don't remember the exact reason. We can check if it's being used in any of the AlgoKit tooling links and potentially remove.

src/features/network/components/network-configs-table.tsx
neilcampbell: Yeah I think there may be. I can check and confirm.

neilcampbell: Also I don't think there is any need to have the "Actions" text visible.

src/features/accounts/pages/account-page.tsx
neilcampbell: Sorry maybe a little more context is useful here.

Your code is doing this:
```
const transformError = (e: Error) => {
  if (is404(e)) {
    const error = new Error(accountInvalidAddressMessage) as StatusError
    error.status = 404
    return error
  }
}
```

You could instead do this:
```
const transformError = (e: Error) => {
  if (is404(e)) {
    e.message = accountInvalidAddressMessage
    return e
  }
}
```

You'd be preserving the thrown error (including the status field) and only mutating the message.

lempira: Oh, I see. yeah, this is much cleaner. Made the change.

src/features/app-interfaces/pages/edit-app-interface-page.tsx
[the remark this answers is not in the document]
lempira: I see what you are saying. I made the change. The only thing is that I had to add the 404 status in the error. Since the useAppInterface call doesn't return a status code, I added one  in order to show the error with the return link

src/features/network/components/network-select.tsx
```diff
@@ -3,37 +3,43 @@ import { Label } from '@/features/common/components/label'
...
-    [location.pathname, navigate, setSelectedNetwork]
+    [currentNetworkId, location.pathname, navigate, setSelectedNetwork]
   )
 
   return (
-    <div className={cn('flex w-48 flex-col')}>
-      <Label htmlFor="network" className={cn('ml-0.5 mb-2')}>
-        Active network
-      </Label>
+    <div className={cn('flex flex-col')}>
```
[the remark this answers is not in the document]
lempira: I added the w-48 constraint back in, but I added it to both the trigger and the content. When I added it to the top-level div as it was before, it didn't have the desired effect. The network name inside the context grew large. I also added a truncate class to the content. The trigger automatically truncates at max w. 
**without truncate**
image

**with truncate**
image

src/features/network/components/network-configs-table.tsx
lempira: I removed the actions header title. I'll wait on confirmation re: meta.className. If there is a need to differentiate the two, perhaps we can use `meta.className` and `meta.headerClassName` although that might be redundant if we need to make them the same most of the time.

src/features/network/components/network-select.tsx
lempira: Oh, I see. Yeah, your're right, we don't need the `setSelectedNetwork` call in the NetworkSelect since the network-page will set the selected network because the URL network and state network will be different. I removed the `setSelectedNetwork` from the handleNetworkChange in the NetworkSelect component.

src/features/app-interfaces/pages/edit-app-interface-page.tsx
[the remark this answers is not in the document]
lempira: I also had to extract the errors message and labels into another file to avoid circular deps. Well, I only had to extract that one const, but I moved all the labels and message in that file to another file.

src/features/network/components/network-select.tsx
neilcampbell: @lempira I initially thought the same, however that only fires when your in a route that has the network in it. For example if you try change the network now from the settings page, it won't work.

lempira: Ugh yeah. I just tried that. I reverted that change. I am not sure what's causing this either. I think the navigate is causing the components to unmount so on remounting in the new page, it's running the useEffect regardless to the deps. Perhaps the process is happening so quickly that selectedNetwork is not yet set in state so it's not returning early like it should. I don't see any perf hits from the double state setting. do you?

neilcampbell: I think that's exactly what's happening.
Yeah agreed, it'll have next to zero performance impact, so I think it's fine.

src/features/network/components/network-select.tsx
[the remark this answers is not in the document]
neilcampbell: I've made some styling tweaks to handle truncate with an ellipses

src/features/network/components/network-configs-table.tsx
neilcampbell: I've checked and this change is all good.

neilcampbell (approved):

Issue #337, opened 2024-11-13
Txn Wizard: Accept keyreg txn arguments from URL query string


lempira:
Objectives
Support for keyreg txn parameters through url

Future compatible for possible extensions to support groups, more transaction types, etc

I would restrict this to just keyregs for the time being, the security UX needs to be considered before we roll this out for everything that may include closeAssetTo, rekey, etc.

Note that rekeyTo field is not specified, make sure not to support it 

Use case
Facilitate Keyreg signing/broadcasting for users of TUI with desktop wallets (Lute, Defly Web, …)

TUI will have a desktop option for signing, which show a lora URL to visit. This should have the keyreg txn params prefilled.

Proposed URL Structure
Base of keyreg arguments: [ARC-78](https://raw.githubusercontent.com/algorandfoundation/ARCs/c4e7561a7dab0d28f14c771881ef883af45f197c/ARCs/arc-0078.md) 

Changes required from ARC-78:

Future proofing: Added index to each argument to allow for groups be passed in the future. E.g. enforce type[0]= instead of accepting type=

We could support both indexes and no indexes but this could be more error prone for devs. Safer to enforce indexes even when it is a single txn.

Add sender= argument

Note: note / xnote arguments are not required for now - feel free to leave them out

Example URL: (online keyreg)

https://lora.algokit.io/transaction-wizard 
  ?type[0]=keyreg
  &sender[0]=I3345FUQQ2GRBHFZQPLYQQX5HJMMRZMABCHRLWV6RCJYC6OO4MOLEUBEGU
  &selkey[0]=-lfw-Y04lTnllJfncgMjXuAePe8i8YyVeoR9c1Xi78c
  &sprfkey[0]=3NoXc2sEWlvQZ7XIrwVJjgjM30ndhvwGgcqwKugk1u5W_iy_JITXrykuy0hUvAxbVv0njOgBPtGFsFif3yLJpg
  &votefst[0]=1300
  &votekd[0]=100
  &votekey[0]=UU8zLMrFVfZPnzbnL6ThAArXFsznV3TvFVAun2ONcEI
  &votelst[0]=11300
  &fee[0]=2000000
Example URL: (offline keyreg)

https://lora.algokit.io/transaction-wizard 
  ?type[0]=keyreg
  &sender[0]=I3345FUQQ2GRBHFZQPLYQQX5HJMMRZMABCHRLWV6RCJYC6OO4MOLEU

PR #338, opened 2024-11-13
Added hook to read and set url search params

lempira:
Closes  issue #337 

I created a hook that will read the search params from the URL and create the transactions in form that is expected by the `defaultTransactions` type in `TransactionsBuilder` Right now, the ticket is only requesting the key registration txn type, but this can be extended for any transaction type in the future and can be used to clone transactions directly from the explorer.

The steps taken in the hook
1. Transform the URL search params to JSON
2. Loop through all the txn and only process the keyreg transactions
3. Transform the URL JSON object to the `BuildKeyRegistrationTransactionResult` type that is expected by the Transaction Builder
4. Validate the KeyReg schema using zod (schema already existed)
5. Pass the transaction through if it validates or pass to an error array. ZodErrors and general errors are currently being handled. Since Zod errors can be gnarly to look at, I only show the path as a string. 
6. Throw an error toast for each txn error. 

**Screenshot of zod errors being thrown**
image


src/features/transaction-wizard/utils/transactions-url-search-params.test.tsx
```diff
@@ -0,0 +1,80 @@
...
+          'votefst[0]': votefst.toString(),
+          'votekd[0]': votekd.toString(),
+          'votekey[0]': votekey,
+          'votelst[0]': votelst.toString(),
+          'fee[0]': fee.toString(),
+        }),
+      })
+      expect(screen.getByText('Online')).toBeInTheDocument()
+      expect(screen.getByText(sender)).toBeInTheDocument()
+      //   expect(screen.getByText(selkey)).toBeInTheDocument()
```
lempira, line 71: These are commented out because what get's sent in the URL params for these values is not what gets rendered. I checked that in the back, the values are the same as the URL params but I can't tell what is changing them in the render.

neilcampbell: I take a look into this.

neilcampbell (approved):

PR #344, opened 2024-11-20
feat: friendly ARC56 error messages

PatrickDinh:
When we can parse the error
image

When we can't parse the error
image

### DIFF
--- a/src/features/abi-methods/utils/parse-errors.ts
+++ b/src/features/abi-methods/utils/parse-errors.ts
@@ -1,5 +1,7 @@
-import { algorandClient } from '@/features/common/data/algo-client'
+import { getApplicationResultAtom } from '@/features/applications/data'
+import { dataStore } from '@/features/common/data/data-store'
 import { BuildableTransactionType, BuildMethodCallTransactionResult, BuildTransactionResult } from '@/features/transaction-wizard/models'
+import { base64ToBytes } from '@/utils/base64-to-bytes'
 import { asError } from '@/utils/error'
 import { AppClient } from '@algorandfoundation/algokit-utils/types/app-client'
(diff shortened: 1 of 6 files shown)


neilcampbell (approved):

src/features/abi-methods/utils/parse-errors.ts
```diff
@@ -0,0 +1,100 @@
...
+  const tealErrorMessage = extractErrorMessage(logicError.message)
+  if (!tealErrorMessage) {
+    return asError(e)
+  }
+
+  return new Error(`Error in transaction ${groupIndex + 1}: ${tealErrorMessage}`)
+}
+
+const extractErrorMessage = (errorString: string): string | undefined => {
+  const regex = /Runtime error when executing .+ in transaction .+: (.+)$/
```
neilcampbell, line 97: I'm wondering if we make this more permissive to potential message changes.
Maybe it's better to simply extract the value after the last `:`
Thoughts?

src/features/abi-methods/utils/parse-errors.ts
```diff
@@ -0,0 +1,100 @@
...
+  const transaction = transactions[groupIndex]
+  if (!transaction || transaction.type !== BuildableTransactionType.MethodCall) {
+    return asError(e)
+  }
+
+  return parseErrorForTransaction(e, groupIndex, transaction)
+}
+
+const parseErrorForTransaction = async (e: unknown, groupIndex: number, transaction: BuildMethodCallTransactionResult) => {
+  const program = await algorandClient.app.getById(BigInt(transaction.applicationId))
```
neilcampbell, line 73: We could instead make this call using our atom cache.

PR #351, opened 2024-11-25
Rm preview app workflow

lempira:
I added this workflow because it appears that deleting all of the preview apps that are generated when a PR is open does not happen automatically on Cloudflare. I have been having to delete them from the dashboard manually. [This ticket has a discussion](https://github.com/cloudflare/pages-action/issues/47)

I tested this on this branch and it worked. I ran it on every PR update and then changed the trigger only to run when PRS are closed.


.github/workflows/remove-preview-apps.yaml
```diff
@@ -0,0 +1,34 @@
+name: Delete Cloudflare Pages Preview Deployment
```
neilcampbell, line 1: Assuming this file needs to be deleted?

neilcampbell (approved):

.github/workflows/remove-preview-apps.yaml
lempira: Yeah, i thought i did a `git mv` to change the file name from `rm-` to `remove-` I meant to just keep the "remove" prefix. I renamed rm to remove and deleted the unused file.

PR #353, opened 2024-11-26
feat: upgrade to use-wallet v3

PatrickDinh:


src/features/network/data/index.ts
```diff
@@ -270,24 +270,6 @@ export const networkConfigAtom = atom<NetworkConfigWithId>((get) => {
   }
   const config = networkConfigs[id]
 
-  // This clears a connected mnemonic wallet after a page reload, as the mnemonic is not stored and cannot be used to sign transactions.
```
PatrickDinh, line 273: @neilcampbell I don't think we need this logic, MNEMONIC wallet is cleared automatically after a page reload. Is there any special cases that I should test?

src/features/wallet/data/selected-kmd-wallet.ts
PatrickDinh, line 1: I split this file to avoid recursive dependencies between `kmd` and `networkConfig`.

src/features/network/data/index.ts
neilcampbell: If that's the case we can likely removed. I'm curious if the stored state is cleared on reload by use-wallet.
3 test cases to run after removing this code:
- connect mnemonic wallet, reload, check you connect the same mnemonic wallet and sign/send a transaction
- connect mnemonic wallet, reload, check you connect a different mnemonic wallet and sign/send a transaction
- connect mnemonic wallet, reload, check you can connect a different wallet provider and sign/send a transaction

PatrickDinh: I can confirm that after reloading, I can connect to a different mnemonic wallet and sign the transactions.

patches/@txnlab+use-wallet+3.10.0.patch
```diff
@@ -0,0 +1,44 @@
+diff --git a/node_modules/@txnlab/use-wallet/dist/index.cjs b/node_modules/@txnlab/use-wallet/dist/index.cjs
```
neilcampbell, line 1: Your fix has been merged in a released, so we can update and remove this patch. Might be worth a quick test after update to confirm.

PR #360, opened 2024-12-04
chore: environment variable substitution for the Tauri build

PatrickDinh:
I can't test the Tauri build fully due to there were no code signing certs setup. I can confirm that the build works all the way to right before code signing.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -75,7 +75,6 @@ jobs:
       release-tag: ${{ steps.create-release-action.outputs.release-tag }}
       release-id: ${{ steps.create-release-action.outputs.release-id }}
 
-  # TODO: VITE_ tokens need to be replaced after building, but before bundling the Tauri app
   build-tauri:
     name: Build Tauri app
     needs:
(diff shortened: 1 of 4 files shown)


.github/workflows/release.yaml
```diff
@@ -84,7 +83,7 @@ jobs:
     strategy:
       matrix:
         # macos-14 is the Apple Silicon M1 runner
-        platform: [ubuntu-20.04, windows-latest, 'macos-12', 'macos-14']
+        platform: [ubuntu-22.04, windows-latest, 'macos-12', 'macos-14']
```
PatrickDinh, line 86: I had to upgrade to ubuntu 22.04 because libwebkit2gtk-4.1-dev package wasn't found in ubuntu 20.04

neilcampbell: Does that alter what ubuntu version we support, assuming going from 20 to 22 results in a different glibc version too. Could an alternative be to install `libwebkit2gtk-4.1` as part of the build process?

neilcampbell: Wait I can see that we are installing it. Is it just not compatible with 20?

neilcampbell (approved):

.github/workflows/release.yaml
PatrickDinh: yes that is the case. Attempting to install `libwebkit2gtk-4.1` on ubuntu 20.04 resulted in a `the package was not found in registry` error. The issue was mentioned here
https://github.com/tauri-apps/tauri/issues/9662#issuecomment-2411428766
Following that issues on GitHub, it seems to be that at least we need Ubuntu 22.04 to build Tauri v2.
----
I tested with a fresh ubuntu 20.04 VM and can confirm that algokit-lora can't be installed 
image

PR #362, opened 2024-12-09
feat: use a custom form to prompt for KMD password and mnemonic

PatrickDinh:
feat: use a custom form to prompt for KMD password and mnemonic

DEMO: 
https://github.com/user-attachments/assets/5326ca30-cc31-4b68-827a-161e7078b0e1


src/features/common/components/promp-form.tsx
```diff
@@ -0,0 +1,44 @@
+import { useCallback } from 'react'
```
neilcampbell, line 1: Filename should be `prompt-form.tsx`

src/features/common/components/wallet-provider.tsx
```diff
@@ -24,9 +38,29 @@ export function WalletProvider({ networkConfig, children }: Props) {
               baseServer: networkConfig.kmd.server,
               token: networkConfig.kmd.token ?? '',
               port: String(networkConfig.kmd.port),
+              promptForPassword: async () => {
+                const password = await openKmdPasswordDialog({ message: 'Enter KMD Password' })
+                if (password === undefined || password == null) {
```
neilcampbell, line 43:
```suggestion
                if (password == null) {
```
That covers the undefined scenario as well.

neilcampbell (approved):

Issue #365, opened 2024-12-13
Incorrect display of large numbers in app global state


uhudo:
Issue: Lora incorrectly displays the last digits of large numbers in the global state of applications, e.g. `33399922244455501` is displayed as `33399922244455500`
It seems to be rounding them.

neilcampbell:
Thanks for reporting @uhudo. Do you have an app id on mainnet or testnet we can inspect?

uhudo:
Unfortunately not. I've just noticed it while doing tests on localnet.

neilcampbell:
No problems, should be pretty easy to reproduce.

PatrickDinh:
hi @uhudo, I have fixed this issue in #366, can you please confirm if it's fixed in your localnet?

uhudo:
Yes, it's correct now.
Thanks for the quick fix!

PR #366, opened 2024-12-13
bug: handle bigint encoding with algod

PatrickDinh:
bug: handle bigint encoding with algod so that bigint can be decoded. This fixes the issue #365 

image


src/features/common/data/algo-client.ts
```diff
@@ -17,13 +17,19 @@ const shouldCreateKmdClient = (config: NetworkConfig) => {
 // Init the network config from local storage
 const networkConfig = settingsStore.get(networkConfigAtom)
 export let indexer = ClientManager.getIndexerClient(networkConfig.indexer)
+
 export let algod = ClientManager.getAlgodClient(networkConfig.algod)
```
neilcampbell, line 21: Should we set this on indexer and kmd as well, just in case?

neilcampbell (approved):

src/features/common/data/algo-client.ts
PatrickDinh: I think that's a good idea, let me do it.

PR #370, opened 2024-12-15
Added a Heatmap

lhcee3:
This a an official adding of a feature called Heatmap, which helps visualize the assets. This feature is an official contribution as part of the Algorand Developer Track Competition. The Heatmap will help visualize, track and analyze wallets on the lora website itself.
Screenshot 2024-12-15 143718


lempira:
Closing this for now. This is not a feature that we would like to add and maintain in Lora at the moment. We have been thinking about building a plugin system that would allow people to create great visualizations like this. But that has been postponed for other features that are needed.

PR #378, opened 2025-01-21
Feature/search result nav link

lempira:
Closes: AK-45

Commit 1: Adds NavLink around the results and use the `result.url` for href.
Commit 2: Fixed that annoying act warning for the `search.test.tsx` file

image

### DIFF
--- a/src/features/search/components/search.test.tsx
+++ b/src/features/search/components/search.test.tsx
@@ -86,19 +86,19 @@ describe('search', () => {
         return executeComponentTest(
           () => render(<Search />, undefined, myStore),
           async (component, user) => {
-            const input = component.getByPlaceholderText(searchPlaceholderLabel)
-            await user.type(input, id)
-            const results = (await component.findAllByText(label, undefined, { timeout: 1000 })).map((result) => result.parentElement)
-            results.forEach((result) => {
-              const link = result!.querySelector('a')
-              const linkText = link?.textContent || ''
-              expect(linkText.substring(0, 3)).toBe(id.substring(0, 3))
-            })
-            const result = results.find((result) => result!.textContent!.includes(type))!
-            await user.click(result)
-
             await waitFor(
               async () => {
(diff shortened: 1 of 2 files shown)


src/features/search/components/search.test.tsx
```diff
@@ -86,14 +86,19 @@ describe('search', () => {
...
+
+                const results = (await component.findAllByText(label, undefined, { timeout: 1000 })).map((result) => result.parentElement)
+                results.forEach((result) => {
+                  const link = result!.querySelector('a')
+                  expect(link).toHaveTextContent(label)
+                })
+
+                const result = results.find((result) => result!.textContent!.includes(type))!
+                await user.click(result)
                 expect(mockNavigate).toHaveBeenCalledWith(`/localnet/${type.toLowerCase()}/${id}`)
```
lempira, line 102: This commit diff is kinda hard to follow. I removed the `waitFor` around the `expect(mockNavigate).toHaveBeenCalledWith(`/localnet/${type.toLowerCase()}/${id}`)` and put it around all the tests. This ensured it handled all the async updates in the test. This is a similar approach @negar-abbasi took in the  `when search results for nfd have been returned` test. I first tried to wrap act statements around each click  event, but that fix seemed cumbersome and I also realized that we are a major version behind the react-testing-library so the act signature was an older one. I'll update dep in another PR.

src/features/search/components/search.tsx
```diff
@@ -93,7 +93,9 @@ export function Search() {
                   {results.map((result) => {
                     return (
                       <CommandItem key={`${result.type}-${result.id}`} value={result.url} onSelect={handleSelection}>
-                        <span>{result.label}</span>
+                        <NavLink className="truncate text-primary underline" to={result.url}>
+                          {result.label}
+                        </NavLink>
```
lempira, line 98: I used a `NavLink` instead of a `TemplatedNavLink` because the URL was already in the `results.url`. I just applied the same styling as `TemplatedNavLink`.

PR #381, opened 2025-01-23
feat: upgrade algosdk + subscriber + utils

PatrickDinh:


src/features/transaction-wizard/mappers/as-description-list-items.tsx
```diff
@@ -518,7 +523,7 @@ const asResourcesItem = (accounts?: string[], assets?: bigint[], apps?: bigint[]
                 <ol className="pl-4">
                   {assets.map((assetId, index, array) => (
                     <li key={index} className="truncate">
-                      <AssetIdLink assetId={Number(assetId)} className="inline text-primary underline">
+                      <AssetIdLink assetId={assetId} className="inline text-primary underline">
                         {Number(assetId)}
```
neilcampbell, line 527:
```suggestion
                        {assetId.toString()}
```
Potentially this is better handling now?
Probably worth doing a search for all the places.

src/utils/random-bigint.ts
```diff
@@ -0,0 +1,5 @@
+import { randomNumber, randomNumberBetween } from '@makerx/ts-dossier'
```
neilcampbell, line 1: I'm assuming this is only used in tests? Probably makes sense to move to src/tests/utils

src/utils/remove-encodable-methods.ts
```diff
@@ -0,0 +1,27 @@
+/* eslint-disable @typescript-eslint/no-explicit-any */
+import { Encodable } from 'algosdk'
+
+export const removeEncodableMethods = <T extends Encodable>(thing: T): Omit<T, 'getEncodingSchema' | 'toEncodingData'> => {
```
neilcampbell, line 4:
```suggestion
export const removeEncodableMethods = <T extends Encodable>(encodable: T): Omit<T, 'getEncodingSchema' | 'toEncodingData'> => {
```

neilcampbell (approved):

PR #382, opened 2025-01-23
fix: support asset transfers for decimal values over the safe number value

neilcampbell:
- Support decimal asset transfer values over the limits of number
- Refine asJson to convert to number where applicable
- Fix zod bigInt schema usage

### DIFF
--- a/src/utils/as-json.ts
+++ b/src/utils/as-json.ts
@@ -24,4 +24,32 @@ export const normaliseAlgoSdkData = (obj: unknown): unknown => {
   return obj
 }
 
-export const asJson = (value: unknown) => JSON.stringify(value, (_, v) => (typeof v === 'bigint' ? v.toString() : v), 2)
+const toNumber = (value: number | bigint) => {
+  if (typeof value === 'number') return value
+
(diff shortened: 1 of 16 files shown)


src/utils/as-json.ts
```diff
@@ -24,4 +24,32 @@ export const normaliseAlgoSdkData = (obj: unknown): unknown => {
...
+  if (value > BigInt(Number.MAX_SAFE_INTEGER)) {
+    throw new Error(`Cannot convert ${value} to a Number as it is larger than the maximum safe integer the Number type can hold.`)
+  } else if (value < BigInt(Number.MIN_SAFE_INTEGER)) {
+    throw new Error(`Cannot convert ${value} to a Number as it is smaller than the minimum safe integer the Number type can hold.`)
+  }
+  return Number(value)
+}
+
+const defaultJsonValueReplacer = (_key: string, value: unknown) => {
+  if (typeof value === 'bigint') {
```
PatrickDinh, line 39: I'm not so sure about `bigint` can be serialised to either string or number ... I think I prefer it to be always string.

PatrickDinh (approved):

PR #389, opened 2025-02-06
feat: Added plausible custom events

lempira:
Added the following plausible analytic events according to [these instructions](https://plausible.io/docs/custom-event-goals):

app-lab-use-existing-app-id
app-lab-deploy-new-app
txn-wizard-add-txn
txn-wizard-simulate-txn
txn-wizard-send-txn
fund-account-form-create-and-fund
fund-account-form-fund


neilcampbell (approved):
Potentially worth confirming these events are captured when running via the Tauri app.

lempira:
> Potentially worth confirming these events are captured when running via the Tauri app.
This is a good point. I am not sure how Plausible works will work with the Tauri app. Will test it out.

PR #400, opened 2025-04-07
chore: move wasm test to the block mapping

neilcampbell:
Moving here means that the mapping is run as part of deconstructing the block that is fetched in the background.
This should help surface any issues quicker.

I also noticed that `closeAmount` doesn't appear to be available in the model.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -5680,9 +5680,9 @@
       }
     },
     "node_modules/caniuse-lite": {
-      "version": "1.0.30001680",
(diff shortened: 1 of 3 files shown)


joe-p:
Nice so I assume this will do the encode/decode on the main page that shows transactions streaming in?

And yeah `closeAmount` is not in `algo_models` since we just care about the models that people will be signing. The `SignedTxnWithAD` fields that you see in a block will be covered by the algod API package.

neilcampbell:
Yeah it runs regardless of the page as it's tied into the block polling code that runs in the background.

> And yeah closeAmount is not in algo_models since we just care about the models that people will be signing
Ah, I see! Makes sense.

PR #406, opened 2025-05-09
Create dependabot.yml

lempira:
Adds dependabot to keep deps up-to-date.


joe-p (approved):
LGTM assuming vitest was updated out of necessity due to an audit issue

lempira:
> LGTM assuming vitest was updated out of necessity due to an audit issue
> 
> Edit: `npm audit` still seems to be failing

Yeah, that was because of the npm audit. I also needed to update vite to the latest minor for npm audit as well. Did that now.

PR #422, opened 2025-05-21
chore: switch to use the new utils-ts

PatrickDinh:


package.json
```diff
@@ -25,7 +25,8 @@
   },
   "dependencies": {
     "@algorandfoundation/algokit-subscriber": "^3.2.0",
-    "@algorandfoundation/algokit-utils": "^9.0.0",
+    "@algorandfoundation/algokit-utils": "^9.2.0-future.3",
+    "@algorandfoundation/tealscript": "https://registry.npmjs.org/@algorandfoundation/tealscript/-/tealscript-0.106.3.tgz",
```
PatrickDinh, line 29: I have to install `@algorandfoundation/tealscript` this way as a walk around for the issue that `algokit-utils` is hosted on GitHub and `tealscript` is hosted on regular npm

.npmrc
```diff
@@ -0,0 +1,3 @@
+@algorandfoundation:registry=https://registry.npmjs.org/
+@algorandfoundation/algokit-utils:registry=https://npm.pkg.github.com/
+//npm.pkg.github.com/:_authToken=${NODE_AUTH_TOKEN}
```
neilcampbell, line 3:
```suggestion
```
Is this needed or does npm find the user .npmrc at `~/.npmrc`?

PatrickDinh: This line isn't needed for GitHub CI/CD because the environment variable is set [here](https://github.com/MakerXStudio/shared-config/blob/main/.github/workflows/node-ci.yml#L99). However, this is used by Cloudflare to set the auth token.

neilcampbell (approved):

PR #425, opened 2025-05-23
feat: adjust how we visually represent transactions from rekeyed accounts

neilcampbell:
Adjust how we visually represent transactions from rekeyed accounts that are initiated by applications.

Simple Before
simple_before

Simple After
simple_after

Complex Before
complex_before

Complex After
complex_after


PatrickDinh (approved):

pbennett:
> @pbennett FYI, this change has been deployed

Huzzah !   🫡

PR #454, opened 2025-06-26
fix: patch algosdk so deleted applications can be fetched

neilcampbell:
Fixes https://github.com/algorandfoundation/algokit-lora/issues/455

When trying to load an application that has been deleted, the following error is returned "Error: Application failed to load". This is due to the algosdk@3 indexer client not handling the null approval and clear program which is returned when the application is deleted.

This PR monkey patches algosdk to convert any null or undefined bytes value it encounters when deserialising JSON into empty bytes value.

An upstream PR to algosdk has been created https://github.com/algorand/js-algorand-sdk/pull/991

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -46,7 +46,7 @@
         "@txnlab/use-wallet": "^4.0.0",
         "@txnlab/use-wallet-react": "^4.0.0",
         "@xstate/react": "^5.0.3",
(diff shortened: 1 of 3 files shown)


neilcampbell:
> TIL 💡 - the patching system

@lempira Yeah `patch-package` is so useful.

PR #456, opened 2025-06-30
feat: add support for viewing application local state

neilcampbell:


src/features/applications/data/application-local-state.ts
```diff
@@ -0,0 +1,79 @@
...
+      return []
+    }
+    const appSpec = await get(createAppSpecAtom(applicationId))
+    return asLocalStateValues(localState.keyValue, appSpec ? asArc56AppSpec(appSpec) : undefined)
+  })
+
+  return [currentEnteredAddressAtom, enteredAddressAtom, resultsAtom] as const
+}
+
+const useSearchAtoms = (applicationId: ApplicationId) => {
```
lempira, line 69: should this be called useApplicationLocalStateSearchAtoms to be more specific what it's search and in case we add search to something else later?

src/features/common/components/data-table.tsx
```diff
@@ -13,10 +13,11 @@ import { useEffect, useState } from 'react'
 import { cn } from '@/features/common/utils'
 import { TableDataContext } from '../../settings/data/table-pagination'
 import { useTablePagination } from '@/features/settings/data/table-pagination'
+import { MESSAGE_TABLE_ROW_DATA_LABEL, NO_RESULTS_TABLE_MESSAGE } from '../constants'
 
 interface DataTableProps<TData, TValue> {
   columns: ColumnDef<TData, TValue>[]
-  data: TData[]
+  data: TData[] | React.ReactNode // Allow for a single ReactNode to be passed, so that errors and loading states can be displayed inline.
```
lempira, line 20: Out of curiousity, why the design choice determining the loading state by passing in a ReactNode. What about using an `loading` and `error` boolean prop. It could leave the data just for the TData and give clearer picture of the different possible states of the component.

src/features/applications/data/application-local-state.ts
neilcampbell: Yeah that's a better name, however it's not exported so only used in this scope. I'll rename though.

src/features/common/components/data-table.tsx
neilcampbell: I like to try design types to not allow invalid states to the representable. In the case where we need to display a message (could be error or info) or loading state, then there would be no data, so allowing data to be set doesn't really makes sense IMO.

We could probably make this clearer by adjusting the data type. We could potentially even use a Loadable<TData[]>, which can be in `hasData`, `hasError` or `loading` states.

neilcampbell: That approach has worked out pretty well, so have changed to that see https://github.com/algorandfoundation/algokit-lora/pull/456/commits/1fb4f093dad03b975e8d3df6f66bcbfe391e12bf

lempira: I like the react-query pattern where the state is separated from data and you test the loading states (loading or error) before you render what's in the data. Better yet, I like using suspense w/ error boundaries so react can handle the loading and error state for you. 
I think having it in the loadable makes sense here, tho.

lempira (approved):

PR #469, opened 2025-07-25
fix: correct the rendering of inner transaction groups

neilcampbell:
Before
image

After
block-52121244-group-_6zYx5gx0Iv3sLRso9Z1Xwr4niLcNyorFpQ_uBVkcBA_ (1)


src/features/common/components/version-display.test.tsx
```diff
@@ -99,7 +99,7 @@ describe('VersionDisplay', () => {
 
       expect(component.getByText(/Built:/)).toBeInTheDocument()
       expect(component.getByText(/Commit:/)).toBeInTheDocument()
-      expect(component.getByText(/1\/15\/2024/)).toBeInTheDocument() // Date formatting
+      expect(component.getByText(RegExp(`${new Date('2024-01-15').toLocaleDateString()}`))).toBeInTheDocument() // Date formatting
```
neilcampbell, line 102: This is so the test is locale agnostic. It failed on my local machine due to Australia format being 15/1/2025

PR #484, opened 2025-08-04
Feat/arc62 circulating supply

p2arthur:
Screenshot 2025-08-04 at 12 26 30 PM

### 📌 Summary

This PR adds support for displaying **Circulating Supply**, **Burned Supply**, and **Reserve Supply** on the ASA asset page for assets conforming to the [ARC-62 standard](https://github.com/algorandfoundation/ARCs/pull/138), using **ARC-3 metadata binding**.

---

### 🧩 Key Changes

- ➕ Added `ARC62` to the `AssetStandard` enum
- 🧠 Implemented `isArc62()` utility to detect ARC-62 conformance via ARC-3 traits
- 🔁 Updated asset mapper and metadata builder to extract:
  - `circulatingSupply`
  - `burnedSupply`
  - `reserveSupply`
- 💡 Displayed these properties on the ASA page when present
- 🛠 Created `executeFundedDiscoveryApplicationCall()` to simulate ARC-62 getter calls using the fee sink address (no signatures needed)
- 🧪 Added unit and integration tests for:
  - ARC-62 detection
  - Onchain query simulation
  - Metadata parsing
- 📦 Included ARC-62 mock assets for testing
- 🧹 Cleaned up unused imports and removed debug logs

---

### 🧪 TestNet Asset Used

- **ARC-3 Example**: `741524548` → Expected Circulating Supply: `1`

---


lempira (commented):
~~I didn't review the tests since you are still working on them~~
I review in my ide but then it froze and  didn't push the comments. Will have to redo.

p2arthur:
You mean the lora application froze?

src/features/assets/components/asset-details.tsx
```diff
@@ -50,6 +53,15 @@ const expandAssetJsonLevel = (level: number) => {
 }
 
 export function AssetDetails({ asset }: Props) {
+  const raw = String(asset.metadata?.arc62Metadata)
+  let parsedArc62 = null
+  try {
+    parsedArc62 = raw ? JSON.parse(raw) : null
+  } catch (e) {
+    parsedArc62 = null
+    throw e
```
lempira, line 62: Is there a reason why you are rethrowing the error here? Do you want the app to error?

src/features/assets/components/asset-details.tsx
```diff
@@ -50,6 +53,15 @@ const expandAssetJsonLevel = (level: number) => {
 }
 
 export function AssetDetails({ asset }: Props) {
+  const raw = String(asset.metadata?.arc62Metadata)
```
lempira, line 56: Instead of casting this as a string, you can use type guard. This will also assign a type to the parsed json

src/features/assets/components/asset-details.tsx
```diff
@@ -85,6 +97,24 @@ export function AssetDetails({ asset }: Props) {
...
+            dt: burnedSupplyLabel,
+            dd: <div>{parsedArc62.burnedSupply}</div>,
+          }
+        : undefined,
+      parsedArc62
+        ? {
+            dt: reserveSupplyLabel,
+            dd: <div>{parsedArc62.reserveSupply}</div>,
+          }
+        : undefined,
```
lempira, line 117: Does Arc62 return all of these values or can some be undefined? If they can be undefined, then you should separate these conditional renders. You can do that if you type the return value as shown above with the type guard. 
Also, i know that using the ternary was the pattern here, but you can achieve the same pattern with 
```
...(parsedARc62.reserveSupply ? [{
            dt: reserveSupplyLabel,
            dd: <div>{parsedArc62.reserveSupply}</div>,
          }] "[]
```

src/features/assets/data/asset-metadata.ts
```diff
@@ -59,6 +62,22 @@ const createAssetMetadataResult = async (
           metadata_url: gatewayMetadataUrl,
           metadata,
         }
+
+        const usesArc62 = isArc62(arc3MetadataResult)
+
+        if (usesArc62) {
+          const arc62AppId = metadata.properties[`arc-62`]['application-id']
```
lempira, line 69: This `arc62AppId` is returning an `any` value. This should be type. You could use another type guard here.

src/features/assets/data/types.ts
```diff
@@ -51,7 +51,12 @@ export type Arc69MetadataResult = {
   }
 }
 
+export type Arc62MetadataResult = {
+  metadata: { circulatingSupply: number; burnedSupply: number; reserveSupply: number }
```
lempira, line 55: Are all of these returned in Arc62? Are any of them optionally returned?

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,114 @@
+import { algorandClient, indexer } from '@/features/common/data/algo-client'
+import algosdk, { getApplicationAddress } from 'algosdk'
+import { Arc3MetadataResult } from '../data/types'
+import { uint8ArrayToUtf8 } from '@/utils/uint8-array-to-utf8'
+
+// Checks if the asset metadata has the ARC-62 property in the correct place
+export const isArc62 = (asset: Arc3MetadataResult): boolean => {
+  if (!asset?.metadata?.properties) return false
+
+  const arc62 = asset.metadata.properties['arc-62'] as { 'application-id'?: number }
```
lempira, line 10: Is there way you can avoid casting here? Could this `asset.metadata.properties` be some union type some where? You can can also type guard. This would avoid runtime errors. Something like
```
function isValidArc62Property(value: unknown): value is { 'application-id': number } {
    return (
      typeof value === 'object' &&
      value !== null &&
      'application-id' in value &&
      typeof (value as any)['application-id'] === 'number'
    )
  }
```

src/features/assets/mappers/asset.ts
```diff
@@ -119,6 +124,11 @@ const asStandardsUsed = (assetResult: AssetResult, metadataResult: AssetMetadata
   if (metadataResult?.arc3?.metadata.properties && isArc16Properties(metadataResult.arc3.metadata.properties)) {
     standardsUsed.add(AssetStandard.ARC16)
   }
+
+  //TODO-ARTHUR: Add conditional to check if the asset follows ARC-62 and execute function to get the circulating supply and burned supply - populate the asset with the data
```
lempira, line 128: Is this done already? I see the conditional check, but not the calls for the other metadata.

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,114 @@
...
+    | number
+    | bigint
+    | string
+    | Uint8Array
+    | Array<boolean | number | bigint | string | Uint8Array>
+    | algosdk.TransactionWithSigner
+    | algosdk.Transaction
+    | Promise<algosdk.Transaction>
+  )[]
+) {
```
lempira, line 97: It looks like you can use `AppClientMethodCallParamsArgs[]` here. I am pretty sure it's the same types. Confirm first.

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,114 @@
...
+  // Using a fee sink address to call the method - simulating will check if caller account has balance
+  const feeSinkAddress = 'Y76M3MSY6DKBRHBL7C3NNDXGS5IIMQVQVUAB6MP4XEMMGVF2QWNPL226CA'
```
lempira, line 99: Does this account for all three environments? The `data/index.ts` file has env constant info.  Also for constants like this, it would be better that they are all uppercase, `FEE_SINK_ADDRESS`

src/utils/funded-discovery.ts
```diff
@@ -0,0 +1,35 @@
+import { algorandClient } from '@/features/common/data/algo-client'
```
lempira, line 1: is this  file duplicated from the `arc62.ts` file?

src/features/assets/components/asset-details.tsx
p2arthur: No, you're right. Shouldn't break the page just because of a Json parsing error. Fixing it right now to just warn and return null

src/features/assets/components/asset-details.tsx
p2arthur: It should always return those since ARC62 is composed by circulating, reserve and burned. I'm composing the metadata by querying the application specified the arc-62 arc3 property.

Interesting pattern! Do you think i Should I keep what we have or implement this new one?

src/features/assets/utils/arc62.ts
p2arthur: Indeed!

src/utils/funded-discovery.ts
p2arthur: Yeah, I forgot to remove this function from the arc62 utils. Moved it to a diff file so we can reuse it

package.json
```diff
@@ -30,7 +30,7 @@
   },
   "dependencies": {
     "@algorandfoundation/algokit-subscriber": "^3.2.0",
-    "@algorandfoundation/algokit-utils": "^9.2.0-future.6",
+    "@algorandfoundation/algokit-utils": "9.1.2",
```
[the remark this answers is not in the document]
p2arthur: Changed because I needed a npm key at the time I started with this repo. I'm gonna change the utils version

src/features/assets/components/asset-details.tsx
lempira: Up to you. Perhaps leave as is. There is a `.filter(isDefined)` that filters out the undefines anyways.

src/features/assets/data/types.ts
lempira: answered in another comment thread

src/features/assets/utils/arc62.ts
lempira: I saw the you removed the cast and just checked if metadata was populated. Is that enough to determining if it's arc-62? What if the metadata wasn't what was expected for arc-62?

src/utils/funded-discovery.ts
```diff
@@ -0,0 +1,26 @@
...
+import { ABIMethod } from 'algosdk'
+
+export async function executeFundedDiscoveryApplicationCall(
+  applicationMethod: ABIMethod,
+  applicationId: bigint,
+  applicationCallArgs?: AppClientMethodCallParamsArgs[]
+) {
+  // Using a fee sink address to call the method - simulating will check if caller account has balance
+  const feeSinkAddress = 'Y76M3MSY6DKBRHBL7C3NNDXGS5IIMQVQVUAB6MP4XEMMGVF2QWNPL226CA'
```
lempira, line 11: I saw that you moved this to this file, but there was a comment about this `finkSinkAddress` I'll. close the other comment and start the thread here

> Does this account for all three environments? The data/index.ts file has env constant info. Also for constants like this, it would be better that they are all uppercase, FEE_SINK_ADDRESS

src/features/assets/utils/arc62.ts
lempira: this code was moved. Opening a thread in the new location.

neilcampbell (commented):
For some reason when I run this branch locally no search results are returned. Do you see the same behaviour?

src/tests/utils/check-arc62.test.ts
```diff
@@ -0,0 +1,60 @@
...
+describe('arc62Utils.isArc62', () => {
+  arc62TestCases.forEach(({ name, metadata, expected }) => {
+    test(name, () => {
+      const assetResult = assetResultMother['testnet-740315456']().build()
+
+      // Extend the assetResult from "asset builder" with the metadata
+      const extended = { ...assetResult, metadata: metadata.metadata }
+
+      // Test if the asset, based on the metadata, is recognized as ARC-62 due to its application-id property checked by isArc62()
+      expect(arc62Utils.isArc62(extended)).toBe(expected)
```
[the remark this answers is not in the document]
p2arthur: Just fixed it by correctly checking for both arc62 property and that the arc62 property uses the correct application-id name.
image

src/features/assets/mappers/asset.ts
```diff
@@ -110,6 +112,9 @@ const asMedia = (assetResult: AssetResult, metadataResult: AssetMetadataResult):
 
 const asStandardsUsed = (assetResult: AssetResult, metadataResult: AssetMetadataResult): AssetStandard[] => {
   const standardsUsed = new Set<AssetStandard>()
+
+  if (!metadataResult || !metadataResult.arc3) return []
```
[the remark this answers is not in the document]
p2arthur: fixed it removing the line and updating the definition of standard used to include arc62 like this.
 
image

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,94 @@
+import { indexer } from '@/features/common/data/algo-client'
+import algosdk, { getApplicationAddress } from 'algosdk'
+import { Arc3MetadataResult, Arc62MetadataResult } from '../data/types'
+import { uint8ArrayToUtf8 } from '@/utils/uint8-array-to-utf8'
+import { executeFundedDiscoveryApplicationCall } from '@/utils/funded-discovery'
+
+// Checks if the asset metadata has the ARC-62 property in the correct place
+export const isArc62 = (asset: Arc3MetadataResult): boolean => {
```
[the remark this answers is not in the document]
p2arthur: Updated to this 
image

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,94 @@
...
+// helper to decode state
+export const decodeAppState = (globalState: algosdk.indexerModels.TealKeyValue[]) => {
+  const decoded: Record<string, string | number | bigint> = {}
+
+  globalState.forEach(({ key, value }) => {
+    const decodedKey = uint8ArrayToUtf8(key)
+
+    if (value.type === 1) {
+      // value.bytes is a Uint8Array representing the address
+      const address = algosdk.encodeAddress(value.bytes)
```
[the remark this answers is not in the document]
p2arthur: Yeah, the example asset includes 4 properties.

I'm not sure how we could filter which properties of the global state to decode in this case

image

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  if (!arc62) return false
+  return arc62 !== undefined
+}
+
+// TODO Arthur - Refactor this - Try to make funded account application discovery more reusable
+export const getArc62CirculatingSupply = async (applicationId: bigint, assetId: bigint) => {
+  const arc62GetCirculatingSupplyMethod = algosdk.ABIMethod.fromSignature('arc62_get_circulating_supply(uint64)uint64')
+
+  const arc62ContractData = await indexer.lookupApplications(applicationId).do()
```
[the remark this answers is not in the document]
p2arthur: I'm not very familiar with those, but I'd love to learn about and implement it.

@lempira what do you think?

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,94 @@
...
+
+  if (!arc62ContractData.application) {
+    throw new Error(`Application with ID ${applicationId} not found.`)
+  }
+
+  const globalAppState = arc62ContractData.application?.params.globalState ?? []
```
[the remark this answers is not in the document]
p2arthur: Yeah, but as of my understanding, its already abstracted in the contract level. To implement the ARC62 spec, you will need the getCirculatingSupply method which handles those cases behind the scenes

src/features/assets/utils/arc62.ts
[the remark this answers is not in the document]
p2arthur: It makes a lot of senseI Will add Jotai in this implementation and make sure to use it on future implementations of the indexer. 

For future reference, are there any other functionalities Jotai could be or currently is useful in Lora @neilcampbell?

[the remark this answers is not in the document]
neilcampbell: Jotai is a global state management library which uses a bottom up approach. As such we use it for any state which is shared throughout the app and encode the fetching/caching of that state within the jotai atoms. It's a super flexible system, which allows you to combine/compose atoms to derive other atoms in a reactive way and build some pretty advanced behaviours.

src/features/assets/utils/arc62.ts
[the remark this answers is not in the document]
neilcampbell: For sure the circulating supply would be accounted for, however I'm more wondering about the addresses that are being shown here. If burned/locked/generic etc addresses are stored box storage, then lora currently would only show them if the contract developer is using global state. Potentially I'm missing something here?

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,101 @@
...
+    throw new Error(`Burned address not found in application ${applicationId} global state.`)
+  }
+
+  const simulateResult = await executeFundedDiscoveryApplicationCall(arc62GetCirculatingSupplyMethod, applicationId, [assetId])
+
+  if (!simulateResult || !simulateResult.returns?.[0]) return
+  return simulateResult.returns[0].returnValue
+}
+
+export const getArc62BurnedSupply = async (applicationId: bigint, assetId: bigint) => {
```
neilcampbell, line 48: These functions to get the specific supply amounts probably shouldn't ever throw, so that it acts like a progressive enhancement where available. If we can't determine the amount simply return undefined.

src/features/assets/utils/arc62.ts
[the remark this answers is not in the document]
neilcampbell: The key field there gives you the value (burned, generic, locked), however if you can use other values or add additional addresses, there isn't much we can do.

I wonder if it's best to try progressively enhance the UI with burned, locked, generic values if available and not support other variation. That means we'll need to be fairly defensive and assume the data isn't always available.

src/utils/funded-discovery.ts
```diff
@@ -0,0 +1,32 @@
...
+import { settingsStore } from '@/features/settings/data'
+import { ABIMethod, Address } from 'algosdk'
+import { AlgorandClient } from '@algorandfoundation/algokit-utils'
+
+export async function executeFundedDiscoveryApplicationCall(
+  applicationMethod: ABIMethod,
+  applicationId: bigint,
+  applicationCallArgs?: AppClientMethodCallParamsArgs[]
+) {
+  let fundedDiscoveryAddress: string | Address = FEE_SINK_ADDRESS
```
neilcampbell, line 13: Did we want to use a different address between MainNet and TestNet?

src/utils/funded-discovery.ts
```diff
@@ -0,0 +1,32 @@
...
+  const networkConfig = settingsStore.get(networkConfigAtom)
+
+  if (!networkConfig.id) return
+  if (networkConfig.id === 'localnet') {
+    const localnetClient = AlgorandClient.defaultLocalNet()
```
neilcampbell, line 18: You should be able to simply import `import { algorandClient } from '@/features/common/data/algo-client'`
Then do:
```
if(await algorandClient.client.isLocalNet()) {
  fundedDiscoveryAddress = await algorandClient.account.localNetDispenser()
}
```

Given that this data is useful for the optional sender feature, it could be even better to encode this network specific funded address inside a jotai atom that can be used anywhere. It'd need to be tied into the network data atom, so when the network is changed, it's updated to the correct address.

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,101 @@
...
+
+  const arc62ContractData = await indexer.lookupApplications(applicationId).do()
+
+  if (!arc62ContractData.application) {
+    throw new Error(`Application with ID ${applicationId} not found.`)
+  }
+
+  const globalAppState = arc62ContractData.application?.params.globalState ?? []
+  const decodedState = decodeAppState(globalAppState)
+  const burnedAddress = decodedState.burned.toString()
```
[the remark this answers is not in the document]
p2arthur: Yeah I think you're right. The burned supply is way too specific of the example arc62 implementation. The spec only requires the implementation of getCirculatingSupply. 

Maybe its better if we keep it to just circulating supply. What you think?

src/features/assets/utils/arc62.ts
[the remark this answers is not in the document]
p2arthur: Was able to implement jotai. Pretty cool stuff
image

src/features/assets/utils/arc62.ts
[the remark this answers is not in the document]
p2arthur: Yeah, following Cosimos advice, the arc62 standard should only display the circulating supply because it's the only required method that needs to be implemented.

Removed references for burned supply and reserve since these are very specific for the implementation of each ARC62 asset.

src/features/assets/utils/arc62.ts
```diff
@@ -0,0 +1,101 @@
...
+  // Fetch appplication data to define both creator and burner addresses in order to populate the method call
+  const arc62ContractData = await indexer.lookupApplications(applicationId).do()
+
+  const globalAppState = arc62ContractData.application?.params.globalState ?? []
+
+  // Decode the global state to get the burned address to populate the method call
+  const decodedState = decodeAppState(globalAppState)
+  const burnedAddress = decodedState.burned.toString()
+
+  const burnedAsset = await indexer.lookupAccountAssets(burnedAddress).assetId(assetId).do()
```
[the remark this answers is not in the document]
p2arthur: Removed burned supply completely

src/utils/funded-discovery.ts
p2arthur: I thought the `FEE_SINK_ADDRESS` was for the fee sink of both testnet and main net.

Will use fee sink for main net and faucet for testnet

image

src/utils/funded-discovery.ts
p2arthur: This was a nice exercise to understand jotai better!

Very clever state management.

image
image

neilcampbell (approved):

PR #515, opened 2025-10-31
fix: convert AVMBytes template parameters from base64 to Uint8Array

### LINKED ISSUE #513: Cannot using an AVMBytes teal template value when deploying an app via app lab
[HelloWorldContract.arc56.json](https://github.com/user-attachments/files/23239960/HelloWorldContract.arc56.json)

### Subject of the issue

When deploying an application using app lab and that application uses an AVMBytes TEAL template variable, the base64 you enter is interpreted as an utf8 string.

The attached ARC56 has an address, AVMBytes and byte[32] TEAL template params.

The decompiled TEAL should show the
ACC_ADDR as bytec_1
ACC_AVM_BYTES as bytec_2
ACC_ARC4_BYTES as bytec_3

bytec_2 should match bytec_3 (if you enter the same base64 value as the input), however they show as below

```
bytec_1 // addr Y76M3MSY6DKBRHBL7C3NNDXGS5IIMQVQVUAB6MP4XEMMGVF2QWNPL226CA
concat
log
intc_0 // 1
return
label4:
bytec_0 // 0x151f7c75
bytec_2 // "x/zNsljw1BicK/i21o7ml1CGQrCtAB8x/LkYw1S6hZo="
concat
log
intc_0 // 1
return
label5:
bytec_0 // 0x151f7c75
bytec_3 // addr Y76M3MSY6DKBRHBL7C3NNDXGS5IIMQVQVUAB6MP4XEMMGVF2QWNPL226CA
concat
```

mrcointreau:
## Fix AVMBytes template parameter base64 encoding

Fixes #513

### Summary

AVMBytes template parameters were incorrectly treated as UTF-8 strings instead of being decoded from base64 to bytes during deployment.

### Solution

Implemented proper base64 ↔ Uint8Array conversion for AVMBytes template parameters, mirroring the existing pattern used for ABI byte array types.

**Key changes:**
- Added `AvmFormItemValue` type for form input values (`string | bigint`)
- Updated `AvmValue` type to include `Uint8Array` for actual deployment values
- Created `avmFormItemValueToAVMValue()`: converts base64 string → Uint8Array for deployment
- Created `asAvmFormItemValue()`: converts Uint8Array → base64 string for form display
- Updated template param mappers to use conversion functions
- Added help text to AVMBytes input fields: "A base64 encoded bytes value"

### DIFF
--- a/src/features/abi-methods/mappers/form-item-mappers.ts
+++ b/src/features/abi-methods/mappers/form-item-mappers.ts
@@ -71,6 +71,7 @@ export const abiTypeToFormItem = (
     return formFieldHelper.addressField({
       label: 'Value',
       field: `${path}` as Path<any>,
+      helpText: 'An Algorand address string',
     })
   }
   if (type instanceof algosdk.ABIArrayDynamicType) {
@@ -131,6 +132,7 @@ export const abiReferenceTypeToFormItem = (formFieldHelper: FormFieldHelper<any>
     return formFieldHelper.addressField({
       label: 'Value',
       field: `${path}` as Path<any>,
+      helpText: 'An Algorand address string',
     })
   }
 
--- a/src/features/abi-methods/mappers/form-item-mappers.ts
+++ b/src/features/abi-methods/mappers/form-item-mappers.ts
@@ -144,6 +144,13 @@ export const avmTypeToFormItem = (formFieldHelper: FormFieldHelper<any>, type: A
       field: `${path}`,
     })
   }
+  if (type === 'AVMBytes') {
+    return formFieldHelper.textField({
+      label: 'Value',
+      field: `${path}`,
+      helpText: 'A base64 encoded bytes value',
+    })
+  }
   return formFieldHelper.textField({
     label: 'Value',
     field: `${path}`,
--- a/src/features/app-interfaces/components/create/deployment-details.tsx
+++ b/src/features/app-interfaces/components/create/deployment-details.tsx
@@ -72,7 +72,7 @@ function FormInner({
         (appInterface) => appInterface.name.toLowerCase() === appInterfaceName.toLowerCase()
       )
       setValue('appInterfaceExists', appInterfaceExists)
-      trigger()
+      trigger('name')
     }
   }, [appInterfaceName, loadableAppInterfaces, setValue, trigger])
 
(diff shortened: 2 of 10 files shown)


neilcampbell (approved):

src/features/abi-methods/mappers/form-item-mappers.ts
```diff
@@ -144,6 +144,13 @@ export const avmTypeToFormItem = (formFieldHelper: FormFieldHelper<any>, type: A
       field: `${path}`,
     })
   }
+  if (type === 'AVMBytes') {
+    return formFieldHelper.textField({
+      label: 'Value',
+      field: `${path}`,
+      helpText: 'A base64 encoded bytes value',
```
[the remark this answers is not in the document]
mrcointreau: Good catches!

Addressed them in b04f9f9.

src/features/app-interfaces/components/create/deployment-details.tsx
```diff
@@ -72,7 +72,7 @@ function FormInner({
         (appInterface) => appInterface.name.toLowerCase() === appInterfaceName.toLowerCase()
       )
       setValue('appInterfaceExists', appInterfaceExists)
-      trigger()
+      trigger('name')
```
neilcampbell: It doesn't look like this is actually triggering on load when the app interface exists.

mrcointreau: Hope I got this right, addressed in 8a4b168

neilcampbell: Doesn't appear to have fixed it. Essentially if you create 2 app interfaces with the same name (can use the same app spec), then on loading `/{network}/app-lab/create`, the name should be validated and if one already exists with that name, a validation error should appear on that field.

neilcampbell: In testing the before state, it appears using `trigger()` didn't originally do this anyway. Let's drop the `trigger()/trigger('name')` entirely.

neilcampbell (approved):

src/features/app-interfaces/components/create/deployment-details.tsx
neilcampbell: Sounds good 👍

PR #516, opened 2025-11-03
Fix/mobile components

p2arthur:
### Summary
Polished layout for latest block/transaction cards and data tables to improve mobile readability and follow the responsive audit plan.

### Changes
- Switched latest cards to **flex layout** (no margin hacks)
- Added **semantic header + content** structure to cards
- Improved table overflow + column visibility on mobile

### Result
Cleaner hierarchy, better small-screen behavior, and consistent UI across dashboard components.

### DIFF
--- a/src/features/common/components/dialog.tsx
+++ b/src/features/common/components/dialog.tsx
@@ -113,7 +113,7 @@ export {
 }
 
 export function MediumSizeDialogBody({ children, className }: { children: React.ReactNode; className?: string }) {
-  return <div className={cn('w-full md:w-[768px] md:max-h-[600px] h-[800px] overflow-auto px-4 pt-1 pb-4', className)}>{children}</div>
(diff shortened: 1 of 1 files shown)


p2arthur:
@neilcampbell 

Should we close the nav side menu, when selecting a search result when in a mobile layout?
I think yes, implemented it

When viewing a transaction (or any other entity) the nav side menu does not show the explore menu item as active.
Fixed this 
image

When adding a application call transaction (or any transaction with a long list of fields) using the transaction wizard, the action buttons are cut off.

You can now scroll down to see the action buttons
image

Mobile layout appears to be cut off on right side and the description lists don't look quite as nice (see screenshot comparing
current prod and these changes)

A few changes to the containers and its overflows did the trick
image

Table header text is top aligned on some tables

Now they're all aligned to the bottom on mobile and middle for other sizes
image

PR #518, opened 2025-11-05
feature: added circle usdc faucet link to testnet dispenser

lempira:
Added Circle TestNet dispenser url to the funding TestNet page

image


neilcampbell:
This will only be visible once the user has signed in to use the TestNet dispenser API.
It feels like this should be an option that doesn't require sign in. I'm wondering if we should rework the UX here?

mrcointreau (approved):
LGTM

PR #566, opened 2026-02-13
Chore/cloudflare prod deploy

p2arthur:
This pull request migrates the deployment of the main branch (beta) from publishing using netlify to publish page on cloud flare instead.

This is part of the testing phase for [this ticket](https://algorandfoundation.atlassian.net/jira/software/c/projects/AK/boards/1054?assignee=712020%3Aca1e0228-c5f4-4dd3-aea7-a38ee08e4d99&selectedIssue=AK-1263)


.github/workflows/release.yaml
```diff
@@ -199,7 +199,7 @@ jobs:
   release-web:
     name: Release Web
     needs: [build-and-release]
-    if: github.ref_name == 'release'
+    if: github.ref_name == 'release' || github.ref_name == 'main'
```
lempira: This is the part where the release web only runs for the release branch. We only want to release the production build on release not on merge to main.

.nsprc
lempira: Why did the notes in these files change?

lempira (changes requested):
Looks good, but the release-web.yml should only be run during the release step, not on main. On Merge to main, there is a preview branch that gets published, and that is enough to be able to preview the betas. The pushes to the release branch are the ones that should trigger the actual release build and publish, the one that will go to production.

.github/workflows/release-web.yaml
```diff
@@ -42,6 +42,10 @@ jobs:
           name: app-build
           path: artifacts/website
 
-      - name: Deploy website to Netlify
+      - name: Deploy to Cloudflare Pages
+        if: github.ref_name == 'release'
```
mrcointreau: @lempira  we already have this guard in `release.yaml` right before this workflow is invoked, it's definitely ok to keep it here as well, wdyt?

mrcointreau (commented):
LGTM, just a minor comment about wrangler integration strategy

lempira (approved):

.github/workflows/release.yaml
p2arthur: Fixed by removing the conditional that runs the publish-docs on main branch.

Now it only runs when pushing to the release branch

.github/workflows/release-web.yaml
```diff
@@ -42,6 +42,10 @@ jobs:
...
         run: |
-          npx netlify-cli@23.7.3 deploy --site ${{ secrets.NETLIFY_SITE_ID }} --auth ${{ secrets.NETLIFY_AUTH_TOKEN }} --dir artifacts/website --prod --no-build
+          npx wrangler@latest pages deploy artifacts/website --project-name=algokit-lora --commit-dirty=true --branch=release
```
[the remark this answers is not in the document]
mrcointreau: Well done! 🚀

